PR #5, opened 2017-07-31
Piper/normalization and validation

pipermerriam:
Adds validation and normalization to all inputs and outputs to ensure that all backends return data in the same format.


pipermerriam:
Also noting that the current approach has now been deemed dumb by me and that I'm going to transition this to just use json schema.  Reinventing the wheel and whatnot.

PR #21, opened 2017-09-28
fix README

changwu-tw:
### What was wrong?

- Parameter needs unicode string.
- typo

### How was it fixed?

- add u prefix to string
- modify get_block_by_numbers --> get_block_by_number


pipermerriam:
Sorry but this is all python 2.  You should go ahead and upgrade to python3 as we won'd be supporting python2 much longer than the end of this year.

changwu-tw:
Ahh I see. One more question, do you consider to use newer version of ethereum like 2.0.4 as the backend?

pipermerriam:
Yes, it's planned to be supported.  That's the main reason for this library is to provide people with choice for their testing backend.

Issue #10, opened 2017-08-03
Write the pyethereum20 backend


pipermerriam:
Finish writing the backend for pyethereum `>=2.0.0,<2.1.0`

carver:
"Finish" might be a strong word, but the basic cases seem to be working, according to the tests. We probably should just add individual issues for specific next steps on pyethereum20 backend needs. Please reopen if I misunderstood something.

PR #26, opened 2017-11-29
[WIP] Pyethereum 2.0 Backend Integration

fubuloubu:
Implements #10

@pipermerriam Please review and comment


pipermerriam:
Things I know need to be changed that may not be obvious.

* add to `extras_require` in setup.py
* add config to `tox.ini` for new backend
* add config in `.travis.yml` for new backend
* add basic docs to `README.md` noting the availability of the `PyEthereum20` backend.

eth_tester/backends/pyethereum/v20.py
```diff
@@ -51,31 +51,45 @@ def get_accounts(self):
...
+        transaction = self.get_transaction_by_hash(txn_hash)
+        state = self.get_state(block_number=transaction.block_hash)
+        return state.receipts
 
     #
     # Account state
     #
     def get_nonce(self, account, block_number=None):
-        raise NotImplementedError("Must be implemented by subclasses")
+        state = self.get_state(block_number)
```
pipermerriam: This looks like you're passing in a block number when the `get_state` function expects a hash?

fubuloubu: Haha, I totally did that.

Untested so far, obviously lol

pipermerriam:
You should look at the base backend class here: https://github.com/fubuloubu/ethereum-tester/blob/411a1490da3bdb16ecc5df296534b24f5f438d69/eth_tester/backends/base.py

* Any methods you find in the current `PyEthereum20` class that aren't in that class can be removed.  
* Any methods that are present in that backend class which aren't in the `PyEthereum20` backend class should be added.

fubuloubu:
Gotcha. Realized `PyEthereum20Backend` is older than `BaseChainBackend` and needs to be updated first

eth_tester/backends/pyethereum/v20.py
fubuloubu: Updated

fubuloubu:
@pipermerriam I believe pyethereum 2.0+ requires python 3.6? Will have to discuss version support, at least match what is listed under [pypi](https://pypi.python.org/pypi/ethereum) for pyethereum itself

pipermerriam:
expanding tests to run under 3.6 is fine.  It's on the roadmap soon anyways.  I wouldn't be surprised if it *just works*.  I'll try.

pipermerriam:
Yup, rebase off of master.  You can run the pyethereum20 tests in a 3.6 only environment (though I think it's worth trying in 3.5)

fubuloubu:
They have 3.5 and 3.6 on pyethereum. Also 2.7. Do we really need/want to support 2.7?

pipermerriam:
You don't need to support 2.7 or 3.4.

pipermerriam:
Looks good but you're going to need to implement the serializers to convert the RLP objects to the expected return types.

See: https://github.com/ethereum/eth-tester/blob/master/eth_tester/backends/pyethereum/v16/serializers.py

It's worth pointing out that `eth-tester` is very strict about the return values, so take a look at the test failures and the v16 serializers to get a sense for the format if things are failing.

pipermerriam:
Also you should think about adopting the folder structure of `v16` as well.

pipermerriam:
Can you rebase this off of master.  There's *stuff* in this PR that shouldn't be like the `.bumpversion` (could have been my fault with one of the commits I tossed in).

eth_tester/backends/pyethereum/v20/main.py
```diff
@@ -0,0 +1,208 @@
...
+        elif fork_name == FORK_STATE_CLEANUP:
+            return self.evm.env.config['CLEARING_FORK_BLKNUM']
+        else:
+            raise UnknownFork("Unknown fork name: {0}".format(fork_name))
+
+    #
+    # Meta
+    #
+    def time_travel(self, to_timestamp):
+        while to_timestamp >= self.get_state().timestamp:
```
pipermerriam: This approach seems problematic as it will have O(n) runtime.  Can you do some investigation to see how we can get down to O(1) by mining a single block with the new timestamp?

eth_tester/backends/pyethereum/v20/main.py
```diff
@@ -0,0 +1,208 @@
...
+
+    #
+    # Mining
+    #
+    def mine_blocks(self, num_blocks=1, coinbase=None):
+        if not coinbase:
+            coinbase=self.get_accounts()[0]
+        block_hashes = []
+        for _ in range(num_blocks):
+            block_hashes.append(
```
pipermerriam: There is a utility in the `eth-utils` library, `eth_utils.to_tuple` which can be used as a decorator on this function to remove this build up of state by just using a yield statement here instead to return the mined blockhash.  Mutability, even benign mutability like this is something I prefer to avoid.

eth_tester/backends/pyethereum/v20/main.py
fubuloubu: I have note to fix this, but I'm still working on it

eth_tester/backends/pyethereum/v20/main.py
fubuloubu: Using now

fubuloubu:
Yeah, there's probably a bunch of crap that ended up in the rebase because I always forget the GitHub way to do the rebase so it recognizes it.

I made a bunch of updates. I actually moved the serializers and utils from v16 forward, I want it so that they both leverage that code, and I just program in the few differences based which version of pyethereum is being used. However, I stumbled into an issue with that approach, I will probably have to move it back again. It's a shame, like 95% of the code is shared.

I got 15/48 tests working, I am having a HUGE problem with all of the rest of the tests (except 1) where this is the error trace I get:
```python
    def mine(self, number_of_blocks=1, timestamp=14, coinbase=a0):
        self.cs.finalize(self.head_state, self.block)
        set_execution_results(self.head_state, self.block)
        self.block = Miner(self.block).mine(rounds=100, start_nonce=0)
>       assert self.chain.add_block(self.block)
E       AssertionError

coinbase   = b'\x82\xa9x\xb3\xf5\x96*[\tW\xd9\xee\x9e\xefG.\xe5[B\xf1'
number_of_blocks = 3
self       = <ethereum.tools.tester.Chain object at 0x7f2df69df080>
timestamp  = 14

../../.pyenv/versions/3.6.3/envs/eth-tester/lib/python3.6/site-packages/ethereum/tools/tester.py:249:
AssertionError
--------------------------------------- Captured stdout setup ---------------------------------------
Initializing chain from provided state
---------------------------------------- Captured stdout call ---------------------------------------
INFO:eth.block  Block pre-sealed, 0 gas used 
INFO:eth.chain  Adding to head head=c6745cf3
INFO:eth.chain  Block 1 (6fbfed80) with parent c6745cf3 invalid, reason: Missing extra data for block
near DAO fork 
```

I believe it has something to do with that last `INFO` printout, basically it refuses to add blocks because of this `extra data` it's looking for. I probably made a mistake with the `set_fork_block()` method where it's not taking. I have no idea

pipermerriam:
I'm going to finish this out via #33 .  Bunch of nasty plumbing work.  Thanks for getting this started.

PR #28, opened 2017-12-01
Fix `setup_tester_chain()`

mhchia:
### What was wrong?
1) The import path of py-evm's `MainnetTesterChain` is wrong.
2) The db generated by `get_db_backend()` is `DB` type, not a `ChainDB`, which has method like `get_state_db()`

### How was it fixed?
1) Correct it to the path of the `MainnetTesterChain` in the latest commit in py-evm master branch.
2) Create a `BaseChainDB` instance from the result of `get_db_backend()`

After fixing these, we can use PyEVMBackend as the backend of `EthereumTester` by
```python
from eth_tester import EthereumTester
from eth_tester.backends.pyevm import PyEVMBackend
t = EthereumTester(PyEVMBackend())
```

#### Cute Animal Picture
23737663_811656992362037_4674797565630460721_o

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -180,7 +180,10 @@ def _execute_and_revert_transaction(chain, transaction, block_number="latest"):
 def _get_vm_for_block_number(chain, block_number, mutable=False):
     block = _get_block_by_number(chain, block_number)
     if mutable and not block.header.is_mutable():
-        block.header.make_mutable()
+        if hasattr(block.header, 'make_mutable'):
+            block.header.make_mutable()
+        else:
+            block.header._mutable = True
     vm = chain.get_vm(header=block.header)
     return vm
 
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -223,7 +223,7 @@ class PyEVMBackend(object):
     def revert_to_snapshot(self, snapshot):
         block = self.chain.get_block_by_hash(snapshot)
         header = self.chain.create_header_from_parent(block.header)
-        self.chain = type(self.chain)(chaindb=self.chaindb, header=header)
+        self.chain = type(self.chain)(chaindb=self.chain.chaindb, header=header)
 
     def reset_to_genesis(self):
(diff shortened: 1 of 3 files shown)


pipermerriam:
I updated the version of py-evm being used in the tests.  Lets see what CI says.

mhchia:
It seems there are still 8 tests failing.
Is this PR helping? or should all the tests be fixed?

pipermerriam:
@mhchia Yes, this is a very helpful PR, are you good fixing the remaining failing tests?

cc @njgheorghita this is what was causing the failures yesterday when we upgraded py-evm versions.

pipermerriam:
@mhchia I did 92482d8 to see if it was just a single line change needed.  Lets see what the test suite says.

pipermerriam:
I pushed a few more commits into your branch fixing some minor failures.  The remaining test failure still needs to be fixed but I don't have time to dig into it now.

mhchia:
@pipermerriam Thank you!
I'm not sure if I can solve the remaining test. But I will try to do it.

mhchia:
@pipermerriam
I looked a little bit more into the tests which fail. It seems they are all related to a fork, and exception thrown from the code below.
In `eth_tester/backends/pyevm/main.py`
```python
def _get_block_by_hash(chain, block_hash):
    block = chain.get_block_by_hash(block_hash)

    if block.number >= chain.get_block().number:
        raise BlockNotFound("No block fuond for block hash: {0}".format(block_hash))

    block_at_height = chain.get_canonical_block_by_number(block.number)
    if block != block_at_height:
        raise BlockNotFound("No block fuond for block hash: {0}".format(block_hash))

    return block
```
All failing tests are due to `block != block_at_height`
If I didn't misunderstand the meaning of `canonical_block`, it seems the check is to make sure that  the block is in the longest chain.

Is it possible to remove this check? or is there any special purpose of the check?
Please correct me if there is anything wrong :) Thank you in advance.

pipermerriam:
@mhchia I just pushed two commits which I think will fix this.  Thanks for doing the other parts of this.  Thanks for the contributoin.

mhchia:
Cool! Thank you so much

PR #29, opened 2017-12-06
 Match filter prefix against longer topic list

carver:
### What was wrong?

Filters were overly aggressive. A log with topics `[A, B]` was not matching filter topics `[A]`.

See https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_newfilter

eg~

> A transaction with a log with topics [A, B]
will be matched by the following topic filters:
> - `[A]` "A in first position (and anything after)"
> - ...

### How was it fixed?

- Fixed a couple of incorrect tests. Added several new ones.
- Only preemptively reject log topics if they are shorter than the filter topics (previously any mismatch)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/core/filter-utils/test_filter_helpers.py
+++ b/tests/core/filter-utils/test_filter_helpers.py
@@ -2,6 +2,10 @@ from __future__ import unicode_literals
 
 import pytest
 
+from eth_utils import (
+    big_endian_to_int,
+    is_list_like,
+)
 from eth_tester.utils.filters import (
     check_single_topic_match,
     check_if_from_block_match,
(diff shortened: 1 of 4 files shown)


tests/core/filter-utils/test_filter_helpers.py
```diff
@@ -21,6 +25,20 @@
 TOPIC_C = b'\x00' * 31 + b'\x02'
 
 
+def topic_to_name(param):
+    if is_list_like(param):
+        return [topic_to_name(t) for t in param]
+    if isinstance(param, bytes):
+        topic_int = big_endian_to_int(param)
+        return chr(topic_int + 65)
```
pipermerriam: very clever

PR #31, opened 2017-12-13
Add `get_code` and `send_raw_transaction`

mhchia:
### What was wrong?
1. `PyEVMBackend.get_code` is not implemented.
2. `EthereumTester.send_raw_transaction` are not implemented. It is useful for `EthereumTester` to provide sufficient functionalities as a provider in`web3.py`.
To make `send_raw_transaction` works, we still need to replace `not_implemented` in the field `'sendRawTransaction'` with `call_eth_tester('send_raw_transaction')`  ( https://github.com/ethereum/web3.py/blob/master/web3/providers/eth_tester/main.py#L211 )

### How was it fixed?
1. Implemented `PyEVMBackend.get_code` referring to `PyEVMBackend.get_nonce`.
2. 
- Implemented `PyEVMBackend.send_raw_transaction`, which `rlp.decode` the `transaction_in_hex` with the transaction sede from `vm.get_transaction_class`. After, calls `chain.apply_transaction` to apply the transaction.
- Implemented `EthereumTester.send_raw_transaction`, which gets the `transaction_hash` returned from `PyEVMBackend.send_raw_transaction` and handles the log filtering works after the transaction is applied.

#### TODO
- [x] Fix failing tests

#### Cute Animal Picture

Cute animal picture


eth_tester/backends/pyevm/main.py
```diff
@@ -366,6 +369,14 @@ def _get_normalized_and_signed_evm_transaction(self, transaction):
         signed_evm_transaction = evm_transaction.as_signed_transaction(signing_key)
         return signed_evm_transaction
 
+    def send_raw_transaction(self, raw_transaction_hex):
+        import rlp
```
pipermerriam: Is there a reason this import isn't at the top level?  If there is can you add a comment so that it's clear why.

pipermerriam:
Can you add test cases for the following.

* Account with no code
* Account with code
* Valid raw transaction
* Invalid raw transaction (fails rlp decoding)

eth_tester/backends/pyevm/main.py
mhchia: No, previously I don't want this import affecting the original codebase, where `rlp` is not used. But you're right it should be at the top level. Thank you I will fix it.

mhchia:
No problem. I'm working on it.

mhchia:
@pipermerriam May I ask a question?
In `EthereumTester.send_raw_transaction`, I'm wondering if it is reasonable to `rlp.decode(raw_transaction_bytes, TransactionClass)`, transform the transaction to the `dict` of transaction, and `self.validator. validate_outbound_transaction(transaction_dict` before calling `EthereumTester.backend.send_raw_transaction(evm_transaction)`? This way we can check if the value in each field in the transaction is valid in advance. But it's an extra overhead. Is it worthy to do it? Thank you!

pipermerriam:
> In `EthereumTester.send_raw_transaction`, I'm wondering if it is reasonable to `rlp.decode(raw_transaction_bytes, TransactionClass)`, transform the transaction to the `dict` of transaction, and do validation with `self.validator.validate_outbound_transaction(transaction_dict)`

I don't think this validation should occur in `EthereumTester` but rather in the backend itself.  Reason being that the actual format of a transaction is likely to change and I'd rather not hard-code that logic into the outer `EthereumTester` class itself.  If you'd like to add that logic at the backend layer I think that is fine.

eth_tester/utils/backend_testing.py
```diff
@@ -221,6 +238,39 @@ def test_mine_multiple_blocks(self, eth_tester):
...
+        raw_privkey = b'\x11' * 32
+        test_key = keys.PrivateKey(raw_privkey)
+        eth_tester.send_transaction({
+            "from": eth_tester.get_accounts()[0],
+            "to": test_key.public_key.to_checksum_address(),
+            "gas": 21000,
+            "value": 1 * denoms.ether,
+        })
+        # create a raw transaction which spends previously the sent funds
+        unsigned_transaction = eth_tester.backend.chain.create_unsigned_transaction(
```
pipermerriam: This isn't going to work since these tests are designed to run across different backends so we can't reach down into the actual backend class to do things since each backend may be different.

What I suggest is hard coding the full signed raw transaction in this test so that you can bypass all of the signing logic.

eth_tester/utils/backend_testing.py
```diff
@@ -190,6 +198,15 @@ def test_get_balance_of_listed_accounts(self, eth_tester):
             assert balance >= UINT256_MIN
             assert balance <= UINT256_MAX
 
+    def test_get_code_account_with_code(self, eth_tester):
+        emitter_address = _deploy_emitter(eth_tester)
+        code = eth_tester.get_code(emitter_address)
+        assert code != '0x'
```
pipermerriam: Would you mind extracting the value of `code` here and adding it to `eth_tester.utils.emitter_contract` so that we can assert the exact bytecode value.

eth_tester/utils/backend_testing.py
mhchia: No problem. Make sense.
I have done it in the new commit.

pipermerriam:
@mhchia just making sure you've seen the test failures.  The `send_raw_transaction` needs to be implemented in the other backends.

mhchia:
@pipermerriam no problem. I will fix them. Thank you for the reminder

mhchia:
@pipermerriam  Regarding to the rest of failing tests in mock backend now, I don't know a good way to support `get_code` and `send_raw_transaction`. Could you give me some suggestions about the following questions?

1. `MockBackend.get_code` fails because no actual vm operation is done in `MockBackend.send_transaction`. I can add logics to insert the code when the parameter `to` of `send_transaction` is `b''`. But this way we are still not sure what to insert as the code(it should be the output from the execution of EVM).
2. `MockBackend.send_raw_transaction`: there is no `Transaction(rlp.Serializable)` logic inside `MockBackend` and it is also weird to add it there(Since here everything is mocked). For now I just get the `transaction_hash` from `raw_transaction_hex` and calls `send_transaction`. And I removed  the check of balance in `test_send_raw_transaction_valid_raw_transaction`, because it seems we don't check balance after applying transactions in the tests in `eth_tester. However, I still haven't come up with a better idea to implement `test_send_raw_transaction_invalid_raw_transaction` without the usage of `rlp.decode(raw_transaction_hex, TransactionClass`, where `MockBackend.send_raw_transaction` to raise `rlp.exceptions.DecodingError`

I really appreciate your help!

pipermerriam:
I suggest skipping tests that require `get_code` and `send_raw_transaction` if `supports_evm_execution == False`.  This should allow you to completely skip implementing these methods in the `MockBackend`.

mhchia:
@pipermerriam I added `skip_if_no_evm_execution` to the tests in MockBackend which I failed to implement.

pipermerriam:
:shipit: whenever.  Ping me and I'll cut you a release when you do.

mhchia:
Thank you for the merging:) Now is it okay that we change `'sendRawTransaction': not_implemented` to `call_eth_tester('send_raw_transaction')` in [`main.py`](https://github.com/ethereum/web3.py/blob/master/web3/providers/eth_tester/main.py#L211) ?

pipermerriam:
Yes, if you'll get a PR open we can get that in pretty quickly.

mhchia:
@pipermerriam Thank you! I have already made the [PR](https://github.com/ethereum/web3.py/pull/520) in web3.py

PR #41, opened 2018-01-07
Fix pyevm backend 

mhchia:
### What was wrong?
Due to the modification of VM class in PyEVM, the structure has changed and some attribute is no longer under `VM`. So any access to those attributes like `PyEVMBackend.get_nonce()` doing will fail.

### How was it fixed?
Change some usage of `vm.xxx` to `vm.state.xxx`

#### Cute Animal Picture

Cute animal picture


mhchia:
Ah the tests are still failing. I will try to solve it tomorrow

PR #42, opened 2018-01-10
make `gas` field optional when estimating gas

carver:
### What was wrong?

It is surprising to require a `gas` field when estimating the amount of gas required.

### How was it fixed?

- Modify the test to try a transaction without gas specified
- (in py-evm) Estimate gas by supplying all available gas in the block, and ignore the specified gas
- Run estimate against pending block instead of latest block, to make sure plenty of gas left in block
- Add a bit of wiggle-room to the estimate (it was too restrictive and causing OOG errors on some web3 tests)
- Enforce a minimum estimate, some scenarios were returning 0 gas used (I guess a failure?)

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -58,6 +58,10 @@ GENESIS_INITIAL_ALLOC = {}
 
 SUPPORTED_FORKS = {FORK_HOMESTEAD, FORK_DAO, FORK_ANTI_DOS, FORK_STATE_CLEANUP}
 
+MINIMUM_GAS_ESTIMATE = 30000
+# A buffer of 1.1 would mean allocate 10% more gas than estimated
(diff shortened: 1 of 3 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -384,15 +388,21 @@ def send_transaction(self, transaction):
         self.chain.apply_transaction(signed_evm_transaction)
         return signed_evm_transaction.hash
 
+    def _max_available_gas(self):
+        header = self.chain.get_block().header
+        return header.gas_limit - header.gas_used
+
     def estimate_gas(self, transaction):
         # TODO: move this to the VM level (and use binary search approach)
```
pipermerriam: https://github.com/ethereum/py-evm/issues/264

PR #43, opened 2018-01-10
Catch py-evm BlockNotFound, reraise as eth-tester BlockNotFound

carver:
### What was wrong?

eth-tester was raising a py-evm exception that web3 didn't recognize: `BlockNotFound`

### How was it fixed?

catch and re-raise the exception as an eth-tester standard `BlockNotFound`

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyethereum/v20/main.py
+++ b/eth_tester/backends/pyethereum/v20/main.py
@@ -73,7 +73,7 @@ def _get_block_by_number(evm, block_number):
 
 def _get_block_by_hash(evm, block_hash):
     block_by_hash = evm.chain.get_block(block_hash)
-    if block_by_hash.number == evm.block.number:
+    if block_by_hash is None or block_by_hash.number == evm.block.number:
         raise BlockNotFound(
             "Block with hash {0} not found on chain".format(block_hash)
         )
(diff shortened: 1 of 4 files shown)


carver:
Py2 compatibility issues.

Will rebase on #45

pipermerriam:
You can rebase on master now.

carver:
Ah, equivalent problem in the other backends, I'll fix those too.

carver:
I think the CI lag actually got worse after moving to the ethereum org queue :(

carver:
Took 4 hours to kick off, 6 min to run :/

carver:
And now that I look at it, I want to change it again, haha.

Basically, something like:

```py
@standardize_exceptions({EVMBlockNotFound: BlockNotFound})
```

So I can avoid the instance attribute.

I might have to do something a little squirrelly like define EVMBlockNotFound = None if pyevm isn't available.

pipermerriam:
ping @jpitts 

Can you double check our travis subscription is active.  Sounds like we might be running at capacity and could maybe benefit from bumping our number of dedicated workers up a bit more.

pipermerriam:
Yes, that API seems better.  Will wait on merge.

PR #46, opened 2018-01-11
Allow calls and transactions to work after a failed call or transaction

jstoxrocky:
### What was wrong?
A failed call or transaction to a contract will prevent all further calls and transactions from working do to a difference in the state-root.

```python
from solc import (
    compile_source,
)
from web3.providers.eth_tester import (
    EthereumTesterProvider,
)
from web3 import (
    Web3,
)

source = (
    """
    contract Test {
        uint public value = 1;
        function setValue(uint _value) {
            value = _value;
        }
        function willThrow() {
            require(false);
        }
    }
    """
)

# Compile contract source
compilation = compile_source(source)
compilation_info = compilation['<stdin>:Test']

# Get Web3
provider = EthereumTesterProvider()
web3 =  Web3(provider)

# Create and deploy contract
contract = web3.eth.contract(
    abi=compilation_info['abi'],
    bytecode=compilation_info['bin'],
    bytecode_runtime=compilation_info['bin-runtime'],
)
txhash = contract.deploy()
receipt = web3.eth.getTransactionReceipt(txhash)
contract = contract(address=receipt['contractAddress'])

# Call works
contract.call().value()
>> 1

# Transaction works
txhash = contract.transact().setValue(2)
contract.call().value()
>> 2

# Throw an error in the contract
try:
    contract.transact().willThrow()
    # contract.call().willThrow() has the same effect
except:
    pass

# Trying to call or transact again throws this error
# txhash = contract.transact().setValue(3) will throw the same error
contract.call().value()
>> VerificationFailed: ('state_root', b'(\x0e\xf5\xc7\xdc\xeb\xfd\x9a\x03\xe5\xc3c\x99d\xaa\ts\xce\x89\xab\xfdM\x03\xd0,\xe9l\x13\xad\xbc\xff\xd8', '==', b'\x115\xc0!\xfb\xfd\xca\xd4\x00\xb3t)\x9c\xd4\x0cHB\xa0\x90\xc7\xdc\xac\xea\xdfN69\x1cS\xcaZ\x18')

```

### How was it fixed?
I added a `try/except/finally` block to the `estimate_gas` function in `eth_tester/backends/pyethereum/v16/main.py` that will revert to the snapshot on a failed transaction. Does this problem exist in `v20`, and `pyevm`? Maybe a fix needs to be provided for those as well.

#### Cute Animal Picture

Cute animal picture


carver:
@jstoxrocky how about adding a test that reproduces the issue into https://github.com/ethereum/eth-tester/blob/master/eth_tester/utils/backend_testing.py ?

If you add the test to a new branch off master, it should fail for pyethereum1.6 (locally running `tox -epy35-pyethereum16`), then be fixed when you `git cherry-pick da7273d` .

It would also answer the question whether the problem exists in other backends.

jstoxrocky:
@carver Thanks for the top about `tox` - never knew what it was until now! I found out that this issue exists only for `ethereum>=1.6.0,<2.0.0`, but I added tests for all version of pyethereum and pyevm anyways. Out of curiosity, is there a recommended backend we should be using for dev work? The README for eth-tester claims that py-evm is in alpha, and apart from this issue, `ethereum>=1.6.0,<2.0.0` seems less janky than `ethereum>=2.1.0,<2.2.0`.

eth_tester/utils/backend_testing.py
```diff
@@ -474,12 +486,112 @@ def test_estimate_gas(self, eth_tester):
...
-        gas_estimation = eth_tester.estimate_gas(dissoc(estimate_call_math_transaction, 'gas'))
+        if not is_pyevm_available():  # py-evm requires the gas parameter
+            estimate_call_math_transaction = dissoc(estimate_call_math_transaction, 'gas')
+        gas_estimation = eth_tester.estimate_gas(estimate_call_math_transaction)
         call_math_transaction = assoc(estimate_call_math_transaction, 'gas', gas_estimation)
         transaction_hash = eth_tester.send_transaction(call_math_transaction)
         receipt = eth_tester.get_transaction_receipt(transaction_hash)
         assert receipt['gas_used'] == gas_estimation
 
+    def test_can_call_after_exception_raised_calling_pyethereum(self, eth_tester):
```
carver: The idea behind this module is that you should be able to write the test once, and it tests across all backends. As much as possible, the caller should be agnostic to which backend is being used.

Anything that doesn't have a unified interface would be a bug.

eth_tester/utils/backend_testing.py
```diff
@@ -474,12 +486,112 @@ def test_estimate_gas(self, eth_tester):
...
+        self.skip_if_no_evm_execution()
+        if is_pyevm_available():
+            pytest.skip('Test only relevant for pyethereum.')
+        elif is_pyethereum16_available():
+            from ethereum.tester import TransactionFailed
+        elif is_pyethereum21_available():
+            from ethereum.tools.tester import TransactionFailed
```
carver: Here's a good example of something that should be unified. These should all be the same exception, not dependent on the back end.

Wait until #43 is merged (just waiting on what I expect to be a green test), then use the same pattern from it to convert internal `TransactionFailed` exceptions into a standard `eth_tester.exceptions.TransactionFailed` exception.

eth_tester/utils/backend_testing.py
```diff
@@ -474,12 +486,112 @@ def test_estimate_gas(self, eth_tester):
...
+        call_value_transaction = _make_call_throws_transaction(
+            eth_tester,
+            throws_address,
+            'value',
+        )
+        raw_result = eth_tester.call(call_value_transaction)
+        result = _decode_throws_result('value', raw_result)
+        assert result == (1,)
+
+    def test_can_call_after_exception_raised_calling_pyevm(self, eth_tester):
```
carver: It is a bug that the py-evm backend doesn't raise the general `TransactionFailed` exception, too. But it may be too long of a fix to include in this PR (I don't know until I take a look later).

If so: you can leave this test as `pyevm`-specific until that bug is fixed. Just comment in this pyevm-specific test that "this is testing **broken** behavior, and when py-evm is fixed, then this test should be deleted, and we should remove the pyevm skip in the general version of the test above."

carver:
> Thanks for the tip about tox - never knew what it was until now!

You bet! I hadn't really used it until recently, either. I like the consistency it usually provides.

> is there a recommended backend I should use for projects involving eth-tester

Not a single global recommendation. I'd say to go with whatever works for you, so pyethereum16 is stable for now. pyethereum2 seems like the worst choice, since it may never reach maturity (essentially all efforts have moved to py-evm). py-evm is a good future-proof choice, but if you want something stable now, it's not there. I'm working on swapping py-evm in for pyethereum in the web3py eth-tester-based tests, so I'll be fixing py-evm-related bugs. But there's still a lot to do.

eth_tester/utils/backend_testing.py
carver: Ok, #43 is merged. You can rebase or merge master to get the convenience decorator from that PR.

pipermerriam:
Sorry I haven't been able to look this over this week, I'll try to cover it first thing next week.

eth_tester/utils/backend_testing.py
```diff
@@ -474,12 +486,112 @@ def test_estimate_gas(self, eth_tester):
             math_address,
             'increment',
         )
-        gas_estimation = eth_tester.estimate_gas(dissoc(estimate_call_math_transaction, 'gas'))
+        if not is_pyevm_available():  # py-evm requires the gas parameter
+            estimate_call_math_transaction = dissoc(estimate_call_math_transaction, 'gas')
+        gas_estimation = eth_tester.estimate_gas(estimate_call_math_transaction)
```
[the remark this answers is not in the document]
jstoxrocky: Roger, thanks for the heads up!

eth_tester/utils/backend_testing.py
jstoxrocky: Sounds good.

jstoxrocky:
@carver thanks for the help!

eth_tester/backends/pyethereum/v16/main.py
```diff
@@ -48,6 +49,17 @@
...
+    replace_exceptions,
+)
+
+
+if is_pyethereum16_available():
+    from ethereum.tester import (
+        TransactionFailed as Pyeth16TransactionFailed,
+    )
+else:
+    Pyeth16TransactionFailed = Exception
```
pipermerriam: Eh, I think this should be set to `None`, otherwise it will replace *all* exceptions with a bare `Exception` which I think will obfuscate errors we don't want to intercept.

eth_tester/backends/pyethereum/v20/main.py
```diff
@@ -46,6 +47,17 @@
...
+    replace_exceptions,
+)
+
+
+if is_pyethereum21_available():
+    from ethereum.tools.tester import (
+        TransactionFailed as Pyeth21TransactionFailed,
+    )
+else:
+    Pyeth21TransactionFailed = Exception
```
pipermerriam: Same comment here about setting to `None` instead.

jstoxrocky: Oh oops, good call.

pipermerriam:
@carver can you give this a final review (since you previously had some requests for change which I think are all cleared up)

carver:
> I'm not a real git expert here. Any pointers on how to fix this?

First, I'd recommend doing a rebase. I find it easier, often, to manage the conflicts on a commit-by-commit basis, rather than a whole branch at once. (even now that I'm comfortable with it, though, I usually do a `git branch saved-pre-rebase` first, so I know I can undo the rebase if I need to)

If you don't have the upstream repo, you can add it with:

```sh
git remote add upstream https://github.com/ethereum/eth-tester.git
git fetch upstream
```

Then rebase with:

```sh
git checkout bug/state_root_after_failed_transaction
git branch pre-rebase

git rebase upstream/master

# follow directions, usually:
# 1. `git status` to see which files are conflicted
# 2. resolve conflicts manually in your editor (looking for <<<<)
# 3. git add the fixed files
# 4. git rebase --continue
# 5. repeat until complete

# happy with how everything went?
git branch -D pre-rebase
```

Travis seems to be hung, but only flake8 is stalled. I tried canceling and restarting with no effect yet. But rebasing is worth doing, regardless of what's happening with travis.

pipermerriam:
replaced by https://github.com/ethereum/eth-tester/pull/48

jstoxrocky:
@carver SUPER helpful! Thanks for all the help! I'll work on my git skills before pushing future PRs!

PR #51, opened 2018-01-19
Now compatible with py-evm alpha 9

carver:
### What was wrong?

py-evm's API changed a lot between alpha 8 and 9, breaking eth-tester.

### How was it fixed?

So much `pdbpp`.

Things I found:
- The public API to apply a transaction to the pending block seems to have gone away! I had to manually set the chain header.
- `vm.execute_transaction()` => `state.execute_transaction()`
- `block.receipts` went away

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -189,9 +189,10 @@ def _get_transaction_by_hash(chain, transaction_hash):
 def _execute_and_revert_transaction(chain, transaction, block_number="latest"):
     vm = _get_vm_for_block_number(chain, block_number, mutable=True)
 
-    snapshot = vm.state.snapshot()
-    computation = vm.execute_transaction(transaction)
-    vm.state.revert(snapshot)
+    state = vm.state
+    snapshot = state.snapshot()
+    computation, _ = state.execute_transaction(transaction)
+    state.revert(snapshot)
     return computation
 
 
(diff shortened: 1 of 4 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -206,6 +207,11 @@ def _get_vm_for_block_number(chain, block_number, mutable=False):
     return vm
 
 
+def _insert_transaction_to_pending(chain, transaction):
```
pipermerriam: Is the term `pending` here meant to refer to the "pending block"?

carver: Yes, I'll expand the name or add a docstring.

pipermerriam: The term *pending* is bugging me because of the collision with "pending transactions".

- A transaction added to the pending block **will** be mined when the block is mined.
- A transaction which is in the pending state, meaning it's in the transaction pool, doesn't have the same guarantees.

Any ideas for different terminology?

carver: I just posted a commit with it renamed to `_insert_transaction_to_pending_block`. Is that sufficient clarity? This is intended to be the first case, so that the transaction will be mined in the next block.

pipermerriam: That is sufficiently descriptive for me.

PR #53, opened 2018-01-24
Upgrade to Py-EVM 0.2.0a10 compatible 

hwwhww:
### What was wrong?
The interface of Py-EVM is changed, so it's not compatible with the latest Py-EVM.

### How was it fixed?
1. Update `eth_tester.backends.pyevm.main._execute_and_revert_transaction`.
2. Update Py-EVM version.
3. In Py-EVM, `BaseChainDB.set_as_canonical_chain_head` was renamed to `BaseChainDB._set_as_canonical_chain_head`, so I also have to change where `eth-tester` calls this function, but it may look not good to call a protected function of py-evm. Do we need to open an issue for that?

#### Cute Animal Picture

norway-1758183_1280

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -191,7 +191,7 @@ def _execute_and_revert_transaction(chain, transaction, block_number="latest"):
 
     state = vm.state
     snapshot = state.snapshot()
-    computation, _ = state.execute_transaction(transaction)
+    computation = state.execute_transaction(transaction)
     state.revert(snapshot)
     return computation
 
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -249,11 +249,11 @@ class PyEVMBackend(object):
     def revert_to_snapshot(self, snapshot):
         block = self.chain.get_block_by_hash(snapshot)
         if block.number > 0:
-            self.chain.chaindb.set_as_canonical_chain_head(block.header)
+            self.chain.chaindb._set_as_canonical_chain_head(block.header)
             self.chain = self.chain.get_chain_at_block_parent(block)
             self.chain.import_block(block)
         else:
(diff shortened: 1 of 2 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -249,11 +249,11 @@ def take_snapshot(self):
     def revert_to_snapshot(self, snapshot):
         block = self.chain.get_block_by_hash(snapshot)
         if block.number > 0:
-            self.chain.chaindb.set_as_canonical_chain_head(block.header)
+            self.chain.chaindb._set_as_canonical_chain_head(block.header)
```
pipermerriam: ewww, do we no longer have a public API we can access for this?  We should fix that (not now but soon) by moving it back to being a public API.

cc @carver to ask what the reason was for making this private?

hwwhww:
@pipermerriam it needs review, thanks!

pipermerriam:
Good to go when CI passes.  I'll get you a release when merged.  Feel free to poke me if it seems like I've forgotten

pipermerriam:
@hwwhww released as `v0.1.0b14`

eth_tester/backends/pyevm/main.py
carver: In some ways, it was to highlight issues like this. The public call should be to [`persist_header_to_db()`](https://github.com/ethereum/py-evm/blob/master/evm/db/chain.py#L262). Giving callers a way to reach into the EVM and violate its own fork choice rule seems like asking for trouble.

In this case, it seems like we should actually be using py-evm's own `snapshot()` and `revert()` here, anyway.

pipermerriam: Hrm, off the top of my head, the py-evm `snapshot()` and `revert()` methods only work within block boundaries.  Need to go recheck this.  If that's the case we need to come up with an alternate solution.

PR #55, opened 2018-01-25
Fix nonce in transaction + tests to cover

monokh:
### What was wrong?

Eth tester (with pyethereum backend at least) did not accept the nonce transaction parameter. 

### How was it fixed?

Allow nonce parameter with inbound transaction and validation.

#### Cute Animal Picture

Cute animal picture


eth_tester/utils/backend_testing.py
```diff
@@ -268,11 +268,28 @@ def test_send_raw_transaction_invalid_raw_transaction(self, eth_tester):
...
+        'test_transaction',
+        (
+            SIMPLE_TRANSACTION,
+            {
+                "to": BURN_ADDRESS,
+                "gas_price": 1,
+                "value": 0,
+                "gas": 21000,
+                "nonce": 1,
+            },
```
pipermerriam: Can you replace this with a constant at the top of the module next to where `SIMPLE_TRANSACTION` is defined.

```python
from cytoolz import assoc

TRANSACTION_WITH_NONCE = assoc(SIMPLE_TRANSACTION, 'nonce', 1)
```

monokh: 👍

monokh:
Seems like pyevm also has this validation issue.

monokh:
Pyevm validates the account nonce against the nonce passed. Would you find it acceptable if i just modify the nonce in the test to be `0` @pipermerriam? This makes the tests pass.

monokh:
Made that change, let me know if you have an issue and we can think of something else

carver:
Hm, I think other nodes let you submit nonces in advance. Maybe we could add an xfail test with a higher nonce, and make sure that `xfail_strict=true` in the pytest.ini

pipermerriam:
@carver I think that may be restricted to raw transactions but I'm guessing.

PR #60, opened 2018-02-10
Handle pending transactions

### LINKED ISSUE #59: Handle pending transactions with a tx pool
* Version: 0.1.0-beta.16
* Python: all
* OS: all

### What was wrong?

Eth tester does not handle pending transactions correctly when disabling auto mining transactions:

Sending a transaction with the same nonce causes an invalid nonce error:

```
from eth_tester import (
    EthereumTester,
    PyEVMBackend,
)

backend = PyEVMBackend()
t = EthereumTester(backend=backend)
t.disable_auto_mine_transactions()
accounts = t.get_accounts()
tx = t.send_transaction({'from': accounts[0], 'to': accounts[1], 'gas': 21000, 'value': 1, 'nonce': 0})
tx2 = t.send_transaction({'from': accounts[0], 'to': accounts[1], 'gas': 21000, 'value': 1, 'nonce': 0})
>>> evm.exceptions.ValidationError: Invalid transaction nonce
```
### How can it be fixed?

The issue is that `send_transaction` and `send_raw_transaction` immediately execute on the vm's state and thereby increment the valid nonce.

### Things to consider for the fix:

- A tx pool needs to be managed at the eth tester level
- It should consider the rules for replacing transactions. e.g. transactions with existing nonce should overwrite #57 
- Will the tx pool work for every backend? 
- `eth-tester` exposes both `send_transaction` and `send_raw_transaction` both of which have different code paths. We need to ensure that we treat them the same as far as the tx pool is concerned.  
- `eth-tester` does not know about the outcome of a transaction (e.g. for figuring out what nonce a transaction will execute at). Could circumvent this by using `get_transaction_by_hash` after executing the transaction.

A very rudimentary fix: 
```
if self.auto_mine_transactions:
    self.mine_block()
else:
    snapshot = self.take_snapshot()
    # If a pending transaction with the same nonce exists, overwrite it?
    self._pending_transactions.append(transaction)
    self.revert_to_snapshot(snapshot)

def enable_auto_mine_transactions(self):
    self.auto_mine_transactions = True
    return [self.send_transaction(tx) for tx in self._pending_transactions]
```

@carver @pipermerriam feel free to modify the issue to clarify things 🙂

monokh:
### What was wrong?

Eth tester does not handle pending transactions correctly when disabling auto mining transactions:

Sending a transaction with the same nonce causes an invalid nonce error when auto mining transactions are disabled.

The issue is that `send_transaction` and `send_raw_transaction` immediately execute on the vm's state and thereby increment the valid nonce.

### How was it fixed?

Implement a transaction pool at the `eth-tester` level so that transactions are only executed when auto mining is enabled again.

#### Cute Animal Picture

Cute animal picture

Fixes #59

### DIFF
--- a/eth_tester/main.py
+++ b/eth_tester/main.py
@@ -72,15 +72,13 @@ def get_default_fork_blocks():
 def handle_auto_mining(func):
     @functools.wraps(func)
     def func_wrapper(self, *args, **kwargs):
-        if not self.auto_mine_transactions:
-            snapshot = self.take_snapshot()
-
-        transaction_hash = func(self, *args, **kwargs)
-
         if self.auto_mine_transactions:
+            transaction_hash = func(self, *args, **kwargs)
             self.mine_block()
         else:
+            snapshot = self.take_snapshot()
             try:
+                transaction_hash = func(self, *args, **kwargs)
                 pending_transaction = self.get_transaction_by_hash(transaction_hash)
                 # Remove any pending transactions with the same nonce
                 self._pending_transactions = remove_matching_transaction_from_list(
@@ -88,7 +86,6 @@ def handle_auto_mining(func):
                 self._pending_transactions.append(pending_transaction)
             finally:
                 self.revert_to_snapshot(snapshot)
-
         return transaction_hash
     return func_wrapper
 
--- a/eth_tester/main.py
+++ b/eth_tester/main.py
@@ -4,6 +4,7 @@ import collections
 import itertools
 import operator
 import time
+import functools
 
 from cytoolz.itertoolz import (
     remove,
@@ -46,6 +47,9 @@ from eth_tester.utils.filters import (
     Filter,
     check_if_log_matches,
 )
+from eth_tester.utils.transactions import (
+    extract_valid_transaction_params,
+)
 
 
 def backend_proxy_method(backend_method_name):
@@ -62,6 +66,29 @@ def get_default_fork_blocks():
         'FORK_ANTI_DOS': 0,
         'FORK_STATE_CLEANUP': 0,
     }
+
+
+def handle_auto_mining(func):
+    @functools.wraps(func)
+    def func_wrapper(self, *args, **kwargs):
+        if not self.auto_mine_transactions:
+            snapshot = self.take_snapshot()
+
+        transaction_hash = func(self, *args, **kwargs)
+
+        if self.auto_mine_transactions:
+            self.mine_block()
+        else:
+            pending_transaction = self.get_transaction_by_hash(transaction_hash)
+            pending_transaction = extract_valid_transaction_params(pending_transaction)
+            # Remove any pending transactions with the same nonce
+            self._pending_transactions = [tx for tx in self._pending_transactions
+                                          if pending_transaction['nonce'] != tx['nonce']]
+            self._pending_transactions.append(pending_transaction)
+            self.revert_to_snapshot(snapshot)
+
+        return transaction_hash
+    return func_wrapper
 
 
 class EthereumTester(object):
@@ -110,6 +137,8 @@ class EthereumTester(object):
     _log_filters = None
     _block_filters = None
     _pending_transaction_filters = None
+
+    _pending_transactions = []
 
     _snapshot_counter = None
     _snapshots = None
@@ -285,6 +314,9 @@ class EthereumTester(object):
     #
     def enable_auto_mine_transactions(self):
         self.auto_mine_transactions = True
+        sent_transaction_hashes = [self.send_transaction(tx) for tx in self._pending_transactions]
+        self._pending_transactions.clear()
+        return sent_transaction_hashes
 
     def disable_auto_mine_transactions(self):
         self.auto_mine_transactions = False
@@ -359,6 +391,7 @@ class EthereumTester(object):
                     raw_log_entry = self.normalizer.normalize_inbound_log_entry(log_entry)
                     filter.add(raw_log_entry)
 
+    @handle_auto_mining
     def send_raw_transaction(self, raw_transaction_hex):
         self.validator.validate_inbound_raw_transaction(raw_transaction_hex)
         raw_transaction = self.normalizer.normalize_inbound_raw_transaction(raw_transaction_hex)
@@ -367,15 +400,10 @@ class EthereumTester(object):
         transaction_hash = self.normalizer.normalize_outbound_transaction_hash(
             raw_transaction_hash,
         )
-
         self._handle_filtering_for_transaction(transaction_hash)
-
-        # mine the transaction if auto-transaction-mining is enabled.
-        if self.auto_mine_transactions:
-            self.mine_block()
-
         return transaction_hash
 
+    @handle_auto_mining
     def send_transaction(self, transaction):
         self.validator.validate_inbound_transaction(transaction, txn_type='send')
         raw_transaction = self.normalizer.normalize_inbound_transaction(transaction)
@@ -396,10 +424,6 @@ class EthereumTester(object):
         )
 
         self._handle_filtering_for_transaction(transaction_hash)
-
-        # mine the transaction if auto-transaction-mining is enabled.
-        if self.auto_mine_transactions:
-            self.mine_block()
 
         return transaction_hash
 
--- a/eth_tester/main.py
+++ b/eth_tester/main.py
@@ -278,13 +278,15 @@ class EthereumTester(object):
         for transaction in self._pending_transactions:
             if transaction['hash'] == transaction_hash:
                 return transaction
+        raise TransactionNotFound(
+            "No transaction found for transaction hash: {0}".format(transaction_hash)
+        )
 
     def get_transaction_by_hash(self, transaction_hash):
         self.validator.validate_inbound_transaction_hash(transaction_hash)
-        pending_transaction = self._get_pending_transaction_by_hash(transaction_hash)
-        if pending_transaction:
-            return pending_transaction
-        else:
+        try:
+            return self._get_pending_transaction_by_hash(transaction_hash)
+        except TransactionNotFound:
             raw_transaction_hash = self.normalizer.normalize_inbound_transaction_hash(
                 transaction_hash,
             )
--- a/eth_tester/main.py
+++ b/eth_tester/main.py
@@ -80,7 +80,6 @@ def handle_auto_mining(func):
             self.mine_block()
         else:
             pending_transaction = self.get_transaction_by_hash(transaction_hash)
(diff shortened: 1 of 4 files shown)


eth_tester/utils/backend_testing.py
```diff
@@ -452,15 +453,15 @@ def test_get_transaction_receipt_for_mined_transaction(self, eth_tester):
...
 
-    def test_get_transaction_receipt_for_unmined_transaction(self, eth_tester):
+    def test_get_transaction_receipt_for_unmined_transaction_raises(self, eth_tester):
         eth_tester.disable_auto_mine_transactions()
         transaction_hash = eth_tester.send_transaction({
             "from": eth_tester.get_accounts()[0],
             "to": BURN_ADDRESS,
             "gas": 21000,
         })
-        receipt = eth_tester.get_transaction_receipt(transaction_hash)
```
monokh: According to [eth_getTransactionReceipt ](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt) this shouldn't work.

Changing this test made me realise that get_transaction_by_hash should be able to retrieve the pending transaction - it doesn't right now.

monokh: It does now

eth_tester/main.py
```diff
@@ -64,6 +68,29 @@ def get_default_fork_blocks():
...
+        transaction_hash = func(self, *args, **kwargs)
+
+        if self.auto_mine_transactions:
+            self.mine_block()
+        else:
+            pending_transaction = self.get_transaction_by_hash(transaction_hash)
+            # Remove any pending transactions with the same nonce
+            self._pending_transactions = [tx for tx in self._pending_transactions
+                                          if not (pending_transaction['nonce'] == tx['nonce']
+                                                  and pending_transaction['from'] == tx['from'])]
```
monokh: and chainId equals?

eth_tester/main.py
```diff
@@ -244,15 +273,25 @@ def get_nonce(self, account, block_number="latest"):
     #
     # Blocks, Transactions, Receipts
     #
+
+    def _get_pending_transaction_by_hash(self, transaction_hash):
+        for transaction in self._pending_transactions:
```
monokh: Not very performant. If that is a concern, can make `_pending_transaction` a dict with key=hash value=transaction

monokh:
If this is ok, docs for `get_transaction_by_hash` and `enable_auto_mine_transactions` will need updating to reflect the change of behaviour

eth_tester/main.py
pipermerriam: Probably, but way less important than the others.

eth_tester/main.py
```diff
@@ -64,6 +68,29 @@ def get_default_fork_blocks():
...
+            snapshot = self.take_snapshot()
+
+        transaction_hash = func(self, *args, **kwargs)
+
+        if self.auto_mine_transactions:
+            self.mine_block()
+        else:
+            pending_transaction = self.get_transaction_by_hash(transaction_hash)
+            # Remove any pending transactions with the same nonce
+            self._pending_transactions = [tx for tx in self._pending_transactions
```
pipermerriam: Can we unroll this comprehension into a small stand alone utility function for better readability as well as easier stand-alone testing.

```python
@to_tuple
def (tx, pending_transactions):
    for transaction in pending_transactions:
        ...
        yield transaction
```

Something that's got the same logic, but unrolled into an easier to read code body.

eth_tester/main.py
pipermerriam: Eh, it'd need to be an ordered dict to preserve transaction order right?  I think that would be fine, but it's not a big deal.  The performance hit here isn't likely to be measurable against the actual evm execution overhead.

eth_tester/main.py
```diff
@@ -244,15 +273,25 @@ def get_nonce(self, account, block_number="latest"):
...
     def get_transaction_by_hash(self, transaction_hash):
         self.validator.validate_inbound_transaction_hash(transaction_hash)
-        raw_transaction_hash = self.normalizer.normalize_inbound_transaction_hash(
-            transaction_hash,
-        )
-        raw_transaction = self.backend.get_transaction_by_hash(raw_transaction_hash)
-        self.validator.validate_outbound_transaction(raw_transaction)
-        transaction = self.normalizer.normalize_outbound_transaction(raw_transaction)
-        return transaction
+        pending_transaction = self._get_pending_transaction_by_hash(transaction_hash)
```
pipermerriam: Rather than having this function return `None` when the transaction isn't found, I think it'd be better if it raised `TransactionNotFound`.  This way nobody *accidentally* returns `None` because they forgot to check the return value.

eth_tester/main.py
carver: I expect the node to reject a transaction if you send it with a different chainid than the node is syncing with, so the handling of failed transactions should be sufficient.

eth_tester/main.py
carver: Not a concern, at least for now. If someone wants to test a really high pending count, we can revisit.

eth_tester/main.py
monokh: Good point, it also makes the flow in `get_transaction_by_hash` easier to read

eth_tester/main.py
```diff
@@ -64,6 +69,28 @@ def get_default_fork_blocks():
...
+
+        if self.auto_mine_transactions:
+            self.mine_block()
+        else:
+            pending_transaction = self.get_transaction_by_hash(transaction_hash)
+            # Remove any pending transactions with the same nonce
+            self._pending_transactions = remove_matching_transaction_from_list(
+                self._pending_transactions, pending_transaction)
+            self._pending_transactions.append(pending_transaction)
+            self.revert_to_snapshot(snapshot)
```
carver: I'm thinking this `revert` might need to be in a try/finally block. Otherwise, if there is any kind of exception after a state change, we will fail to roll back.

eth_tester/utils/transactions.py
```diff
@@ -0,0 +1,31 @@
+# TODO: Should this be moved to a common package like eth-utils?
```
carver: Yeah, `VALID_TRANSACTION_PARAMS` seems to have popped up in several places already, so worth talking about extracting it. eth-utils feels a little wrong, nothing else in there is quite so tightly bound to chain implementation details as "which fields are in a transaction". Maybe we could add a `Transaction` rlp class to https://github.com/ethereum/eth-rlp and have `VALID_TRANSACTION_PARAMS` be something like `Transaction.fields`.

eth_tester/main.py
```diff
@@ -285,6 +326,10 @@ def get_transaction_receipt(self, transaction_hash):
     #
     def enable_auto_mine_transactions(self):
         self.auto_mine_transactions = True
+        sent_transaction_hashes = [self.send_transaction(extract_valid_transaction_params(tx))
```
carver: Can we be strict here and _reject_ transactions that have invalid transaction params?

eth_tester/main.py
monokh: I'm not sure I understand. Isn't `revert_to_snapshot` doing the rollback? And if that fails, is there really a meaningful way to continue? 

It feels to me like the whole function should completely cleanly otherwise the final state is not really usable

eth_tester/utils/transactions.py
monokh: Sounds good. I thought maybe the utility functions could also go somewhere. Would like to tackle this as a separate issue as both `eth-tester` and `web3.py` need changed.

eth_tester/main.py
monokh: I think this is too much of un unlikely scenario to be handled. These transactions are added to the list from `get_transaction_by_hash` which _should_ be valid. The only real purpose of this extraction is to not send the params like v,r,s from the dict that is returned by `get_transaction_by_hash`.

We could add validation when we add the transaction to the list but I don't really see the purpose.

eth_tester/main.py
carver: Sorry, what I meant was something like:

```py
snapshot = self.take_snapshot()

try:
    transaction_hash = func(self, *args, **kwargs)
    pending_transaction = self.get_transaction_by_hash(transaction_hash)
    ...
    self._pending_transactions.append(pending_transaction)
finally:
    self.revert_to_snapshot(snapshot)
```
```

File: eth_tester/main.py (hunk 8 above)
Comment (carver) [reply eth_tester/main.py]: > These transactions are added to the list from `get_transaction_by_hash`

Fair enough, I'm okay skipping this validation.

File: eth_tester/main.py (hunk 6 above)
Comment (monokh) [reply eth_tester/main.py]: Oh 😄 I see what you mean. That's it updated.

File: eth_tester/main.py [hunk 9]
```diff
@@ -64,6 +69,30 @@ def get_default_fork_blocks():
     }
 
 
+def handle_auto_mining(func):
+    @functools.wraps(func)
+    def func_wrapper(self, *args, **kwargs):
+        if not self.auto_mine_transactions:
+            snapshot = self.take_snapshot()
+
+        transaction_hash = func(self, *args, **kwargs)
```
carver: I know this starts to make the function more awkward, but I think we should also revert if `func` to raise an exception too.

Maybe we should just have two main blocks, whether `auto_mine` is on, and just duplicate the `func()` call, like:

```py
+        if self.auto_mine_transactions:
+            transaction_hash = func(self, *args, **kwargs)
+            self.mine_block()
+        else:
+            snapshot = self.take_snapshot()
+            try:
+                transaction_hash = func(self, *args, **kwargs)
+                pending_transaction = self.get_transaction_by_hash(transaction_hash)
+                # Remove any pending transactions with the same nonce
+                self._pending_transactions = remove_matching_transaction_from_list(
+                    self._pending_transactions, pending_transaction)
+                self._pending_transactions.append(pending_transaction)
+            finally:
+                self.revert_to_snapshot(snapshot)
```

I find that easier to follow than the repeated "if auto-mine" tests.

monokh: Ah yes that definitely is easier to read 😄

eth_tester/utils/transactions.py
pipermerriam: Note that we shouldn't call it `Transaction` because we're on the cusp of having multiple disparate transaction types with different fields.

pipermerriam:
@monokh lemme know if your ready for me to merge/release this.

monokh:
@pipermerriam Yes, please merge and release :) then I can update https://github.com/ethereum/web3.py/pull/586 to use the new version. I tested locally and all the failing tests were resolved with these 'eth-tester' changes

PR #61, opened 2018-02-12
Allow call method to query past state

bweick:
### What was wrong?
In order to implement comprehensive tests for https://github.com/ethereum/web3.py/pull/552 we needed to add functionality to `eth_tester` that allowed one to query past chain state using the `call` method.

### How was it fixed?
A block_number parameter was passed to the `_execute_and_revert_transaction` function and then also passed through to `_get_normalized_and_signed_evm_transaction` function in order to make sure the nonces matched so that the call method's "transaction" could be executed and then reverted.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/utils/backend_testing.py
+++ b/eth_tester/utils/backend_testing.py
@@ -65,7 +65,6 @@ from .throws_contract import (
     _make_call_throws_transaction,
     _decode_throws_result,
 )
-from eth_tester.backends.pyethereum.utils import is_pyethereum16_available
 
 
 PK_A = '0x58d23b55bc9cdce1f18c2500f40ff4ab7245df9a89505e9b1fa4851f623d241d'
@@ -490,7 +489,6 @@ class BaseTestBackendDirect(object):
         result = _decode_math_result('add', raw_result)
         assert result == (20,)
 
-    @pytest.mark.skipif(is_pyethereum16_available(), reason="v1.6 not supported")
     def test_call_query_previous_state(self, eth_tester):
         self.skip_if_no_evm_execution()
 
--- a/eth_tester/utils/backend_testing.py
+++ b/eth_tester/utils/backend_testing.py
@@ -490,6 +490,8 @@ class BaseTestBackendDirect(object):
         assert result == (20,)
 
     def test_call_query_previous_state(self, eth_tester):
+        self.skip_if_no_evm_execution()
+        
         math_address = _deploy_math(eth_tester)
         call_math_transaction = _make_call_math_transaction(
             eth_tester,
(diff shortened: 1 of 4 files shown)


pipermerriam:
Can you look into why tests are failing in CI?

bweick:
@pipermerriam Was some formatting, but the bigger issue is that querying anything but the latest block isn't supported by either `pyethereum16` or `pyethereum21` so the tests applied to those evms are inherently going to fail (whereas they will be fine on pyevm). Let me know how you want me to proceed.

carver:
I think the best solution is to implement it for the backends.

---

For pyethereum21, the tester chain's `call()` looks like a good template, specifically `apply_message` on an old state: https://github.com/ethereum/pyethereum/blob/v2.1.5/ethereum/tools/tester.py#L202-L211

To get the state at the block of interest, check `chain.mk_poststate_of_blockhash`

---

For pyethereum16, I'm not sure. This seems like the place to start, but I don't see anything convenient for getting the state of old blocks, and applying a transaction to them: https://github.com/ethereum/pyethereum/blob/v1.6.1/ethereum/tester.py#L268

---

In summary: maybe implement pyethereum21, since it seems within reach, and investigate pyethereum16 for a bit. If not sure, you could mark the test as `xfail` for that backend.

bweick:
Implemented `pyethereum21` but seems like a bit of a chore (if at all possible) to do `pyethereum16`. Could attempt to recreate `chain.mk_poststate_of_blockhash` in the backend but would really like to try to get the original web3 issue that drove this dive into `eth-tester` out instead, feels like it has higher value since I would imagine most people have moved on from using `pyethereum16`.

pipermerriam:
I'm ok with pyethereum16 not supporting this functionality.

eth_tester/utils/backend_testing.py
```diff
@@ -489,9 +490,10 @@ def test_call_add(self, eth_tester):
         result = _decode_math_result('add', raw_result)
         assert result == (20,)
 
+    @pytest.mark.skipif(is_pyethereum16_available(), reason="v1.6 not supported")
```
pipermerriam: Hrm, instead of skipping this here, I think it needs to be skippend in the pyethereum16 specific subclass of this testing class.  Otherwise, this test will get skipped if pyethereum16 is installed at all, even if I'm testing the `py-evm` backend.

bweick:
@pipermerriam think we might be waiting on your approval to merge this. Thanks!

carver:
You addressed his only comment, and he had a couple days, so I'll go ahead and merge.

PR #63, opened 2018-02-21
Replace left_pad with rjust

dylanjw:
### What was wrong?
left_pad was removed from eth-utils.

### How was it fixed?
replaced left_pad with rjust built-in method.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/mock/factory.py
+++ b/eth_tester/backends/mock/factory.py
@@ -6,17 +6,17 @@ from cytoolz.dicttoolz import (
 )
 
 from eth_utils import (
-    keccak,
-    to_dict,
+    apply_to_return_value,
+    is_bytes,
     is_dict,
     is_integer,
     is_list_like,
-    is_text,
-    is_bytes,
     is_null,
-    force_bytes,
+    is_text,
+    keccak,
+    to_bytes,
+    to_dict,
     to_tuple,
-    apply_to_return_value,
 )
 
 from eth_tester.utils.accounts import (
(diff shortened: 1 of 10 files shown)


pipermerriam:
The py-evm ones should be fixed with https://github.com/ethereum/py-evm/pull/380

Still need to fix the pyethereum ones.

dylanjw:
Should `force_bytes` be removed as well?

carver:
> Should force_bytes be removed as well?

Yes

carver:
Looks good so far

pipermerriam:
Also, `eth-utils` in setup.py should be `eth-utils>=1.0.0b2,<2.0.0`.

You can also pin all dependencies within a major version while you're there if they are not already.

pipermerriam:
`eth-keys` should also be bumped to the latest beta release.

dylanjw:
@pipermerriam @carver I think Im getting a little off track.  Maybe you could orient me a bit?

the function `stringify` eth_tester/backends/mock/factory.py uses force_bytes in a couple places:

```
@apply_to_return_value(b'|'.join)
@to_tuple
def stringify(value):
    if is_bytes(value):
        yield value
    elif is_text(value):
        yield force_bytes(value)
    elif is_list_like(value):
        yield b''.join((
            b'(',
            b','.join((stringify(item) for item in value)),
            b')',
        ))
    elif is_dict(value):
        yield b''.join((
            b'{',
            b','.join((
                b":".join((stringify(key), stringify(item)))
                for key, item
                in value.items()
            )),
            b'}',
        ))
    elif is_integer(value):
        yield force_bytes(str(value))
    elif is_null(value):
        yield 'None@{0}'.format(id(value))
    else:
        raise TypeError("Unsupported type for stringification: {0}".format(type(value)))
```

`stringify` is used later in the mock/factory.py to process values being fed into keccack().

Should I be updating `stringify`, or make changes further up the call stack such that no values need processing at this point, so stringify can be removed?

carver:
`stringify` is a confusing name for this. It's more like `bytes_repr`.

I think modifying the implementation there is ok.

```py
    elif is_text(value):
        yield force_bytes(value)
```

to:
```py
    elif is_text(value):
        yield to_bytes(text=value)
```

and:
```py
    elif is_integer(value):
        yield force_bytes(str(value))
```

to:
```py
    elif is_integer(value):
        yield to_bytes(text=str(value))
```

_I don't know why it's converting int to str and then to bytes, but this is how to write such a thing. If all tests succeed with `yield to_bytes(value)` in the last code block, then I would do that instead._

PR #64, opened 2018-02-22
Carver/pyethereum support

carver:
### What was wrong?

A few issues in the pyethereum backend since the eth-utils b2 release.

`remove_0x_prefix` was being called with `bytes` input, and was unnecessary.
`decode_abi` was being called with hexstr input, but only bytes are valid now.

### How was it fixed?

Based on #63 

- remove all prefix calls
- decode hex to bytes before decoding abi
_(It kind of seems like eth-tester should be returning bytes instead of a hexstr, but I don't want to introduce another breaking change in the middle of this cluster)_

#### Cute Animal Picture

Cute animal picture


carver:
Blocking on #380 or #389

PR #67, opened 2018-02-23
Contract txns

### LINKED ISSUE #66: Mismatch on the acceptable input/output for contract creation transactions
* Version: 0.1.0-beta.20
* Python: 3.6

### What was wrong?

This is most noticeable when posting a pending transaction and then mining it later:

- You post the pending transaction that creates a contract (with no `to` key)
- When retrieving the pending transaction to add to the block, the normalizer returns the transaction with `to` set to `b''`
- The inbound normalizer/validator chokes on a `to` value that is not a normal address

cc @mhchia 

### How can it be fixed?

Since eth-tester generally follows the practice of strict requirements on both inputs and outputs, one way to solve this would be:

1. Specify an empty `to` field by either:
  a. leaving it out of the transaction
  b.  including a `'to'` key in the transaction dict, with value: `''`
2. Normalization should update to include
  a. inbound normalization should normalize `''` to `b''`
  a. inbound normalization should normalize a missing key to `b''`
  b. outbound normalization should normalize from `b''` to `''`
3. ~When saving pending transactions to mine, do not apply outbound normalization first~ _It's ok as is, because the transaction gets inbound normalized when applying to the block_

carver:
Fixes #66

Problem & solution discussed here ^

#### Cute Animal Picture

Cute animal picture


carver:
I'm not sure why pyethereum21 is failing. I'll have to pick it up tomorrow.

eth_tester/utils/transactions.py
```diff
@@ -8,7 +8,7 @@
     'from',
     'to',
     'gas',
-    'gasPrice',
+    'gas_price',
     'value',
     'data',
     'nonce',
```
mhchia: Will that make sense to remove `chainId` in the next line? 
It seems not the param of transaction, at least for now.

carver: Thanks @mhchia 

I *suspect* this was trying to address the same issue as https://github.com/ethereum/web3.py/pull/656 which is now merged, so I think you're right that this can be removed.

mhchia: Okay I see. I didn't know that issue.
Thank you for pointing it out

mhchia:
@carver 
I just tested with this branch and tried to find the problem with the failing tests occurred in our sharding branch in py-evm.
In eth-tester, it seems we still need to handle the case when a transaction sent with raw_transaction(i.e. calls `send_raw_transaction`) is inside the pending block in manually mining mode.
In this case, since the transaction is already well-constructed and signed, there might not be the corresponding key in `PyEVMBackend`, therefore the `PyEVMBackend._key_lookup` will fail.

```
self = <eth_tester.backends.pyevm.main.PyEVMBackend object at 0x11007da90>
transaction = {'data': b'`\x005`\x1cRt\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00` Ro\x7f\x...3', 'from': b'~s\x05\xb8\\\xf8\x03\xa4\xcf\x0c\x90_T\xcf\x1f\xf1\x92\xa8\xe6\xbe', 'gas': 3000000, 'gas_price': 1, ...}
block_number = 'latest'

    def _get_normalized_and_signed_evm_transaction(self, transaction, block_number='latest'):
>       signing_key = self._key_lookup[transaction['from']]
E       KeyError: b'~s\x05\xb8\\\xf8\x03\xa4\xcf\x0c\x90_T\xcf\x1f\xf1\x92\xa8\xe6\xbe'

~/eth_tester/backends/pyevm/main.py:397: KeyError
```

I'm still thinking how to solve it more elegantly.
My current thought options are:
1. Add a flag for each transaction in pending transaction pool, to indicate whether it is raw transaction or normal transaction dict. Besides, in `_add_transaction_to_pending_block` check if which type a transaction is, and respectively use `send_transaction` or `send_raw_transaction`
2. Change all transaction to raw_transaction(i.e. completely signed transaction in hex or `rlp.Serializable`). But it requires changes like adding API in `PyEVMBackend`.
Do you have any suggestion on this?
Thank you so much

carver:
Ok, I have a test in place. My current favorite option is to add a `raw` key to the transaction dict (like parity does) when returning it from `get_transaction_by_hash`. Then, the transaction can be re-sent later without having to sign it again.

mhchia:
Thank you for the help.
I can keep working on this.
Will update if I have progress

mhchia:
@carver 
I have change the transaction dicts passed to `PyEVMBackend.send_transaction` to the full transaction dict which contains `v`, `r`, `s` in the following commit
https://github.com/mhchia/eth-tester/commit/4a44494599123186d882bae159d3c08cad1895ef
It makes the `pyevm` tests run but I think will cause other backends to fail.
Just want some feedback if this kind of fix is applicable. If yes, I will further fix the other fails.
Thank you:)

carver:
Cool, I pulled in your commit, and added a few more pieces. This is ready for review to merge.

---

_Side note about readability_

I liked the succinctness of this line, but left it out of my changes while I decide if it's more or less readable:
```py
if {'r', 's', 'v'} <= set(transaction.keys()):
```

I think your line might be better, but it doesn't read like English in my head yet, the same way `all(key in transaction for key in {'r', 's', 'v'})` does. It might just be a matter of getting used to it and polling for other people's style preferences.

pipermerriam:
@carver and @mhchia I think this is what you're looking for.

```python
if {'r', 's', 'v'}.issubset(transactin.keys()):
    # do thing
```

mhchia:
@carver Really appreciate your work! The code is a lot cleaner now.
And thank you for indicating the readability issue. I don't think my line is readable either.
And I really prefer the one @pipermerriam mentioned. It is really readable.

carver:
> And I really prefer the one @pipermerriam mentioned. It is really readable.

I agree. Updated the PR, and I think it's ready to go, after a :+1: from a review.

eth_tester/main.py
```diff
@@ -463,7 +470,15 @@ def _add_transaction_to_pending_block(self, transaction):
             if is_locked:
                 raise AccountLocked("The account is currently locked")
 
-        raw_transaction_hash = self.backend.send_transaction(raw_transaction)
+        if {'r', 's', 'v'}.issubset(transaction.keys()):
+            if hasattr(self.backend, 'send_signed_transaction'):
```
pipermerriam: If you go ahead and add a stub implementation of the `send_signed_transaction` to the `eth_tester.backends.base.BaseChainBackend` you can instead use a `try/except` here, catching the `NotImplementedError` which feels a bit cleaner.

eth_tester/utils/backend_testing.py
```diff
@@ -264,8 +264,10 @@ def test_mine_multiple_blocks(self, eth_tester):
     #
     @pytest.mark.parametrize('is_pending', [True, False])
     def test_send_raw_transaction_valid_raw_transaction(self, eth_tester, is_pending):
-        if is_pending and not hasattr(eth_tester.backend, 'send_signed_transaction'):
-            pytest.xfail("backend does not support presigned pending transactions")
+        if is_pending:
+            from eth_tester.backends import PyEthereum16Backend, PyEthereum21Backend
```
pipermerriam: 👍 

I think we're right around the corner from being able to deprecate those backends.  They were originally written to provide a smooth upgrade path for populus, but it's not looking like that is actually going to happen (needing a smooth upgrade path) since populus is looking like it's going to require some significant gutting to get back to a nice state at which point forcing people onto the new py-evm based backend is not really a big deal anymore.

Nothing actionable other than going forward we should probably just accept failures in the pyethereum backends and mark them as xfails or something.

eth_tester/utils/backend_testing.py
```diff
@@ -264,8 +264,10 @@ def test_mine_multiple_blocks(self, eth_tester):
...
+            if isinstance(eth_tester.backend, (PyEthereum16Backend, PyEthereum21Backend)):
+                pytest.xfail("backend does not support presigned pending transactions")
```
pipermerriam: Can you add `xfail=strict` to the `pytest.ini` for this project so that these and any other `xfail` tests don't accidentally pass.

pipermerriam: Blarg, no big deal.  My internet connection must have dropped when I initially submitted this comment.

PR #71, opened 2018-03-03
Move estimate_gas to vm level in py-evm backend

dylanjw:
depends on: https://github.com/ethereum/py-evm/pull/432

### What was wrong?

eth-tester implemented `estimate_gas` before py-evm.

### How was it fixed?

This change delegates the py-evm backends `estimate_gas` to py-evm.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -48,9 +48,13 @@ from .utils import is_pyevm_available
 if is_pyevm_available():
     from evm.exceptions import (
         BlockNotFound as EVMBlockNotFound,
+        InvalidInstruction as EVMInvalidInstruction,
+        Revert as EVMRevert,
     )
 else:
     EVMBlockNotFound = None
+    EVMInvalidInstruction = None
+    EVMRevert = None
 
 
 ZERO_ADDRESS = 20 * b'\x00'
@@ -457,6 +461,9 @@ class PyEVMBackend(object):
         header = self.chain.get_block().header
         return header.gas_limit - header.gas_used
 
+    @replace_exceptions({
+        EVMInvalidInstruction: TransactionFailed,
+        EVMRevert: TransactionFailed})
     def estimate_gas(self, transaction):
         evm_transaction = self._get_normalized_and_unsigned_evm_transaction(
             dict(
@@ -464,8 +471,7 @@ class PyEVMBackend(object):
                 gas=self._max_available_gas(),
                 nonce=0))
 
-        estimate = self.chain.estimate_gas(evm_transaction)
-
+        return self.chain.estimate_gas(evm_transaction)
 
     def call(self, transaction, block_number="latest"):
         # TODO: move this to the VM level.
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -47,7 +47,7 @@ from .utils import is_pyevm_available
 
 if is_pyevm_available():
     from evm.exceptions import (
-        BlockNotFound as EVMBlockNotFound,
+        HeaderNotFound as EVMHeaderNotFound,
         InvalidInstruction as EVMInvalidInstruction,
         Revert as EVMRevert,
     )
@@ -55,7 +55,7 @@ if is_pyevm_available():
         SpoofTransaction as EVMSpoofTransaction
     )
 else:
-    EVMBlockNotFound = None
+    EVMHeaderNotFound = None
     EVMInvalidInstruction = None
     EVMRevert = None
 
@@ -359,13 +359,13 @@ class PyEVMBackend(object):
     #
     # Chain data
     #
-    @replace_exceptions({EVMBlockNotFound: BlockNotFound})
+    @replace_exceptions({EVMHeaderNotFound: BlockNotFound})
     def get_block_by_number(self, block_number, full_transaction=True):
         block = _get_block_by_number(self.chain, block_number)
         is_pending = block.number == self.chain.get_block().number
         return serialize_block(block, full_transaction, is_pending)
 
-    @replace_exceptions({EVMBlockNotFound: BlockNotFound})
+    @replace_exceptions({EVMHeaderNotFound: BlockNotFound})
     def get_block_by_hash(self, block_hash, full_transaction=True):
         block = _get_block_by_hash(self.chain, block_hash)
         is_pending = block.number == self.chain.get_block().number
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -9,6 +9,10 @@ from cytoolz import (
 
 import rlp
 
+from evm.utils.spoof import (
+    SpoofTransaction
+)
+
 from eth_utils import (
     encode_hex,
     int_to_big_endian,
@@ -422,6 +426,11 @@ class PyEVMBackend(object):
         if 'to' not in transaction:
             yield 'to', b''
 
+    def _get_normalized_and_unsigned_evm_transaction(self, transaction, block_number='latest'):
+        normalized_transaction = self._normalize_transaction(transaction, block_number)
+        evm_transaction = self.chain.create_unsigned_transaction(**normalized_transaction)
+        return evm_transaction
+
     def _get_normalized_and_signed_evm_transaction(self, transaction, block_number='latest'):
         signing_key = self._key_lookup[transaction['from']]
         normalized_transaction = self._normalize_transaction(transaction, block_number)
@@ -453,10 +462,20 @@ class PyEVMBackend(object):
         return header.gas_limit - header.gas_used
 
     def estimate_gas(self, transaction):
-        signed_evm_transaction = self._get_normalized_and_signed_evm_transaction(
+        evm_transaction = self._get_normalized_and_unsigned_evm_transaction(
             dict(transaction, gas=self._max_available_gas()),
         )
-        return self.chain.estimate_gas(signed_evm_transaction)
+
+        spoof_evm_transaction = SpoofTransaction(
+            evm_transaction,
+            get_sender=lambda: transaction['from'],
+            sender=transaction['from'],
+            intrinsic_gas=21000,
+            s=1,
+            r=1,
+            v=37)
+
+        return self.chain.estimate_gas(spoof_evm_transaction)
 
     def call(self, transaction, block_number="latest"):
         # TODO: move this to the VM level.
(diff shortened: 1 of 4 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -5,6 +5,10 @@
 
 import rlp
 
+from evm.utils.spoof import (
+    SpoofTransaction
+)
```
carver: This import has to be moved inside the `is_pyevm_available()` check.

eth_tester/backends/pyevm/main.py
```diff
@@ -419,18 +427,20 @@ def _max_available_gas(self):
...
-        computation = _execute_and_revert_transaction(self.chain, signed_evm_transaction, 'pending')
-        if computation.is_error:
-            raise TransactionFailed(str(computation._error))
-
-        gas_used = computation.gas_meter.start_gas - computation.gas_meter.gas_remaining
+        spoof_evm_transaction = SpoofTransaction(
+            evm_transaction,
+            get_sender=lambda: transaction['from'],
+            sender=transaction['from'],
+            intrinsic_gas=21000,
```
carver: This probably shouldn't be hard-coded, in case future hard forks change it. Can't we just leave it as the default from whatever `evm_transaction` has?

eth_tester/backends/pyevm/main.py
```diff
@@ -419,18 +427,20 @@ def _max_available_gas(self):
...
+            s=1,
+            r=1,
+            v=37)
 
-        return int(max(gas_used * GAS_ESTIMATE_BUFFER, MINIMUM_GAS_ESTIMATE))
+        return self.chain.estimate_gas(spoof_evm_transaction)
```
carver: I guess a longer term solution is that py-evm should do something similar about spoofing the transaction _inside_ `estimate_gas()`

eth_tester/utils/backend_testing.py
```diff
@@ -646,7 +646,11 @@ def test_estimate_gas(self, eth_tester):
         call_math_transaction = assoc(estimate_call_math_transaction, 'gas', gas_estimation)
         transaction_hash = eth_tester.send_transaction(call_math_transaction)
         receipt = eth_tester.get_transaction_receipt(transaction_hash)
-        assert receipt['gas_used'] == gas_estimation
+        assert receipt['gas_used'] <= gas_estimation
+        # Tolerance set to the default py-evm tolerance:
+        # https://github.com/ethereum/py-evm/blob/f0276e684edebd7cd9e84cd04b3229ab9dd958b9/evm/estimators/gas.py#L77
+        # https://github.com/ethereum/py-evm/blob/f0276e684edebd7cd9e84cd04b3229ab9dd958b9/evm/estimators/__init__.py#L11
+        assert receipt['gas_used'] >= gas_estimation - 21000
```
carver: :+1:

tests/backends/test_pyevm.py
```diff
@@ -25,4 +44,23 @@ def eth_tester():
...
+        self.skip_if_no_evm_execution()
+
+        throws_address = _deploy_throws(eth_tester)
+        call_will_throw_transaction = _make_call_throws_transaction(
+            eth_tester,
+            throws_address,
+            'willThrow',
+        )
+        with pytest.raises(InvalidInstruction):
+            eth_tester.estimate_gas(dissoc(call_will_throw_transaction, 'gas'))
```
carver: We shouldn't be exposing internal py-evm exceptions through eth-tester. The exceptions raised should be the same no matter the back end. Looks like we need to define a new exception, then convert it. Here is an example: https://github.com/ethereum/eth-tester/blob/master/eth_tester/backends/pyevm/main.py#L331

tests/backends/test_pyevm.py
```diff
@@ -25,4 +44,23 @@ def eth_tester():
...
+
+        call_set_value_transaction = _make_call_throws_transaction(
+            eth_tester,
+            throws_address,
+            'setValue',
+            fn_args=(2,),
+        )
+        gas_estimation = eth_tester.estimate_gas(dissoc(call_set_value_transaction, 'gas'))
+        assert gas_estimation
```
carver: Can we assert more about this result?

eth_tester/backends/pyevm/main.py
```diff
@@ -419,18 +427,20 @@ def _max_available_gas(self):
         return header.gas_limit - header.gas_used
 
     def estimate_gas(self, transaction):
-        # TODO: move this to the VM level (and use binary search approach)
-        signed_evm_transaction = self._get_normalized_and_signed_evm_transaction(
+        evm_transaction = self._get_normalized_and_unsigned_evm_transaction(
```
carver: While pinned to the py-evm alpha 11, we should probably set `block_number='pending'` for gas estimates.

tests/backends/test_pyevm.py
dylanjw: Should py-evm change to indicate an error has occurred with the `estimate_gas`, without exposing the exception?

eth_tester/backends/pyevm/main.py
dylanjw: This PR was started before https://github.com/ethereum/py-evm/pull/454, so I removed the SpoofTransaction here.

eth_tester/backends/pyevm/main.py
dylanjw: I don't grok.  I thought all transactions are `pending`.  `block_number` isnt a transaction parameter. :space_invader:

carver: It's okay, I was misremembering something about how that worked. It only is used to change the nonce if it's not present. No change needed.

tests/backends/test_pyevm.py
carver: Not exactly sure what you're saying. I think we should add a new `eth_tester.exceptions.InvalidInstruction` (or something like it), and in the py-evm backend for `estimate_gas()`, we should capture the py-evm `InvalidInstruction` and raise the eth_tester one.

dylanjw: I get it now.  I rebased and broke that line reference.  I added InvalidInstruction and Revert (:balloon: :baloon: Byzantium :balloon::balloon:)

eth_tester/backends/pyevm/main.py
```diff
@@ -466,19 +476,17 @@ def _max_available_gas(self):
...
-            dict(transaction, gas=self._max_available_gas()),
-        )
-
-        computation = _execute_and_revert_transaction(self.chain, signed_evm_transaction, 'pending')
-        if computation.is_error:
-            raise TransactionFailed(str(computation._error))
-
-        gas_used = computation.get_gas_used()
+        evm_transaction = self._get_normalized_and_unsigned_evm_transaction(
+            dict(
```
pipermerriam: awkward indentation... maybe try:

```
evm_transaction = self._get_normalized_and_unsigned_evm_transaction(dict(
    transaction,
    gas=self._max_available_gas(),
    nonce=0,
))
```

eth_tester/exceptions.py
```diff
@@ -28,3 +28,11 @@ class AccountLocked(Exception):
 
 class TransactionFailed(Exception):
     pass
+
+
+class Revert(Exception):
```
pipermerriam: Should these be subclasses of `TransactionFailed`?  Or maybe just use `TransactionFailed` anyplace that you are currently using `Revert`?  I think my preference is for the later.

tests/backends/test_pyevm.py
```diff
@@ -11,6 +15,17 @@
...
 
+from eth_tester.utils.math_contract import (
+    _deploy_math,
+    _make_call_math_transaction,
+    _decode_math_result,
+)
+from eth_tester.utils.throws_contract import (
+    _deploy_throws,
+    _make_call_throws_transaction,
+    _decode_throws_result,
```
pipermerriam: All of these imports (and this diff in general) suggest that these are not used and can be removed?

tests/backends/test_pyevm.py
```diff
@@ -25,4 +40,4 @@ def eth_tester():
 
 
 class TestPyEVMBackendDirect(BaseTestBackendDirect):
-    pass
+        pass
```
pipermerriam: Bonus indentation?

eth_tester/backends/pyevm/main.py
```diff
@@ -466,19 +474,17 @@ def _max_available_gas(self):
...
+                transaction,
+                gas=self._max_available_gas(),
```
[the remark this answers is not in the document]
dylanjw: :+1: I'll test that and take it out.

eth_tester/backends/pyevm/main.py
```diff
@@ -466,19 +474,17 @@ def _max_available_gas(self):
...
+                nonce=0))
```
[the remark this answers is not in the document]
dylanjw: This is a bad bandaid.  I need to add an override for the nonce in the py-evm implementation of "do_call".  I'll make a pull request.

tests/backends/test_pyevm.py
```diff
@@ -7,10 +7,6 @@
     PyEVMBackend,
 )
 
-from eth_tester.backends.pyevm.utils import (
```
palango: Why do you remove this. This seems necessary to run the tests.

dylanjw:
Im getting an error in an unrelated (?) test:

tests/backends/test_pyevm.py::TestPyEVMBackendDirect::test_get_block_missing <- eth_tester/utils/backend_testing.py FAILED

```
    def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
        """
            Returns the requested block header as specified by block hash.
    
            Raises HeaderNotFound if it is not present in the db.
            """
        validate_word(block_hash, title="Block Hash")
        try:
            header_rlp = self.db[block_hash]
        except KeyError:
            raise HeaderNotFound(
>               "No header with hash {0} found".format(encode_hex(block_hash))
            )
E           evm.exceptions.HeaderNotFound: No header with hash 0x0000000000000000000000000000000000000000000000000000000000000000 found

block_hash = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
self       = <evm.db.chain.ChainDB object at 0x7f2f79aa89e8>
```

dylanjw:
This just needs the py-evm version bump.

pipermerriam:
`py-evm==v0.2.0a17` is ready for you.

pipermerriam:
I went ahead and did the bump, lets see if tests pass.

dylanjw:
Tests passed. @pipermerriam Ok to merge?

pipermerriam:
@dylanjw lemme give it one more pass (tonight or first thing tomorrow) and then we can merge.

pipermerriam:
@dylanjw ready for a release with this in it or is there more that needs to happen?

PR #79, opened 2018-04-03
remove fuzz tests

pipermerriam:
### What was wrong?

I am a little to in love with property based testing and originally wrote some fuzz testers for the backends.  They mostly aren't used and are not really useful.

### How was it fixed?

Removed them.

#### Cute Animal Picture

bn-gy930_catfly_g_20150215224256


pipermerriam:
Most of them haven't been running and the ones that were were low value *IMHO*.  I'm game to re-add them but I'd like to at least spend some more effort thinking about what types of fuzz tests are of value.

PR #82, opened 2018-04-13
Update v signature validation to support 37 and 38

jzvelc:
### What was wrong?
V value validation in signature supported only 0, 1, 27 and 28 values. When using latest Web3.py raw transactions have v set to 37/38 so validation fails.

### How was it fixed?
Adding 37 and 38 as valid values resolves this issue.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/validation/outbound.py
+++ b/eth_tester/validation/outbound.py
@@ -9,6 +9,7 @@ from eth_utils import (
 )
 
 from eth_tester.constants import (
+    UINT256_MAX,
     UINT2048_MAX,
 )
 from eth_tester.exceptions import (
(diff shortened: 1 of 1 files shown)


carver:
Cool, thanks for identifying the problem, and going the extra mile of digging into the code. I'd say let's go all the way and support [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md) which means essentially any `v >= 35 and v <= UINT_256_MAX` is also potentially valid, where `UINT_256_MAX` is : `(2 ** 256) - 1`.

jzvelc:
@carver I updated this PR, can you have a look.

pipermerriam:
Also, because I used the web editor, I appear to have introduces a linting error.  Would you mind cleaning it up when you write the requested tests?

pipermerriam:
Feel free to revert to the `in range(...)` check.  Well done python3.

jzvelc:
@carver I rebased to master and excluded @pipermerriam's commit. I think we can merge this.

PR #91, opened 2018-05-10
Insert gas value when creating tx for gas estimation

dylanjw:
Its a required field for transaction objects

### What was wrong?
Gas estimation fails if a gas parameter is not supplied.

### How was it fixed?
Insert a value for gas when creating the evm transaction.

#### Cute Animal Picture

Cute animal picture


dylanjw:
This is the last change blocking https://github.com/ethereum/web3.py/pull/649.

PR #92, opened 2018-05-16
Refact default fork block

PabloLefort:
Cute animal picture

### DIFF
--- a/eth_tester/backends/pyethereum/v16/main.py
+++ b/eth_tester/backends/pyethereum/v16/main.py
@@ -189,16 +189,16 @@ class PyEthereum16Backend(BaseChainBackend):
     def get_supported_forks(self):
         return SUPPORTED_FORKS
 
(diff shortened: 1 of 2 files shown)


pipermerriam:
Yeah, what you've done here is not behaviorally the same as the previous code (due to the case that @carver posted)

PabloLefort:
Yup, we saw the same. But, just trying..
```python
>>> def a(i):
...     print(i or 0)
... 
>>> a(0)
0
>>> a(1)
1
>>> a(None)
0
```

carver:
But this PR works more like:

```py
def printer(my_num=0):
    print(my_num)

>>> printer(0)
0
>>> printer(1)
1
>>> printer(None)
None
>>> printer(my_num=None)
None
```

PabloLefort:
Yes. I'll revert this.
Thanks either!

PR #93, opened 2018-05-17
WIP: transaction status

PabloLefort:
Add `status` attribute when calling `get_transaction_receipt` only when settted Byzantium fork

### DIFF
--- a/eth_tester/backends/pyethereum/v20/main.py
+++ b/eth_tester/backends/pyethereum/v20/main.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+import os
+
 import pkg_resources
 
 from semantic_version import (
@@ -242,6 +244,7 @@ class PyEthereum21Backend(BaseChainBackend):
             self.evm.chain.env.config['CLEARING_FORK_BLKNUM'] = fork_block or 0
         elif fork_name == FORK_BYZANTIUM:
             self.evm.chain.env.config['METROPOLIS_FORK_BLKNUM'] = fork_block or 0
+            os.environ['ETHEREUM_TESTER_VALIDATOR'] = 'eth_tester.validation.byzantium.ByzantiumValidator'
         else:
             raise UnknownFork("Unknown fork name: {0}".format(fork_name))
 
@@ -376,6 +379,19 @@ class PyEthereum21Backend(BaseChainBackend):
             transaction_hash,
         )
         state = _get_state_by_block_hash(self.evm, block.hash)
+
+        if self.get_fork_block('FORK_BYZANTIUM') is not None:
+            status = int.from_bytes(state.receipts[transaction_index].state_root,
+                                    byteorder='big')
+            return serialize_transaction_receipt(
+                block,
+                transaction,
+                state.receipts[transaction_index],
+                transaction_index,
+                is_pending,
+                status=status,
+            )
+
         return serialize_transaction_receipt(
             block,
             transaction,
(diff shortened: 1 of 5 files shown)


eth_tester/backends/pyethereum/v20/main.py
```diff
@@ -376,6 +379,19 @@ def get_transaction_receipt(self, transaction_hash):
             transaction_hash,
         )
         state = _get_state_by_block_hash(self.evm, block.hash)
+
+        if self.get_fork_block('FORK_BYZANTIUM') is not None:
```
pipermerriam: Shouldn't this be compared to the current block number?  In a chain where Byzantium is enabled but not yet active, this would try to push a `status` field into the receipt.

eth_tester/utils/backend_testing.py
```diff
@@ -315,6 +315,37 @@ def test_send_transaction(self, eth_tester, test_transaction):
...
+        ),
+        ids=[
+            'Simple transaction',
+            'Transaction with nonce',
+            'Create Contract - empty to',
+            'Create Contract - missing to',
+        ],
+    )
+    def test_get_transaction_receipt_byzantium(self, eth_tester, test_transaction):
+        if eth_tester.backend.__class__.__name__ == 'PyEthereum16Backend':
```
pipermerriam: `if isinstance(eth_tester.backend, PyEthereum16Backend)`

though I believe the pattern for skipping tests on specific backends is not to do it here in the main test module but rather here: 

https://github.com/PabloLefort/eth-tester/blob/218b66eb87d542d28bfdf3546846997df9220e54/tests/backends/test_pyethereum16.py#L32-L34

eth_tester/utils/backend_testing.py
```diff
@@ -315,6 +315,37 @@ def test_send_transaction(self, eth_tester, test_transaction):
...
+        backend.set_fork_block(FORK_BYZANTIUM, 0)
+        byzantium_eth_tester = eth_tester.__class__(backend=backend)
+        accounts = byzantium_eth_tester.get_accounts()
+        assert accounts, "No accounts available for transaction sending"
+
+        transaction = assoc(test_transaction, 'from', accounts[0])
+        txn_hash = byzantium_eth_tester.send_transaction(transaction)
+        txn = byzantium_eth_tester.get_transaction_receipt(txn_hash)
+
+        assert txn['status']
```
pipermerriam: This should probably be `assert 'status' in txn`.

tests/backends/test_pyethereum21.py
```diff
@@ -23,6 +24,8 @@
 def eth_tester():
     if not is_pyethereum21_available():
         pytest.skip("PyEthereum >=2.0.0,<2.2.0 not available")
+    if os.environ.get('ETHEREUM_TESTER_VALIDATOR') is not None:
+        del os.environ['ETHEREUM_TESTER_VALIDATOR']
```
pipermerriam: you should use the [`monkeypatch`](https://docs.pytest.org/en/latest/monkeypatch.html) fixture to accomplish this.

PabloLefort:
Thanks for the feedback @pipermerriam i'll do some refact

eth_tester/utils/backend_testing.py
carver: Or `assert 'status' in txn and txn['status'] == 1` if we know the transaction is successful.

PabloLefort:
thanks @pcppcp

PR #96, opened 2018-05-18
Tweak readme so rst output for pypi is valid

carver:
### What was wrong?

The project description was not rendering on pypi.

### How was it fixed?

Hunting down invalid rst formatting with a lot of `restview --long --pypi`, and tweaking the markdown so it can be converted correctly.

Also a couple bonus link target fixups.

_Do we even need all these manual targets? Targets are automatically generated at the headers, so the value seems minimal_

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,7 @@ The `fork_name` parameter follows the same restrictions as `set_fork_block`
 The chain can only time travel forward in time.
 
 <a id="api-time_travel"></a>
+
 #### `EthereumTester.time_travel(timestamp)`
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
Ok, so not a build problem but a formatting problem.  Does `restview` allow us to do this in CI as part of our linting run?

carver:
I don't see an obvious way. :(

`restview` is launches a page in a local web server, and displays the error or successfully rendered page in browser.

PR #100, opened 2018-06-06
Add `status` field to Byzantium tx receipts

pcppcp:
### What was wrong?
Receipts in `eth-tester` don't include `status` field. In Raiden, it prevents us from using `eth-tester` backend for our tests/CI.

### How was it fixed?
Status field was added for transactions  that occur after `FORK_BYZANTIUM`. I've used some ideas discussed in #93 

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/serializers.py
+++ b/eth_tester/backends/pyevm/serializers.py
@@ -77,7 +77,13 @@ def serialize_transaction(block, transaction, transaction_index, is_pending):
     }
 
 
-def serialize_transaction_receipt(block, receipts, transaction, transaction_index, is_pending):
+def serialize_transaction_receipt(
+    block,
+    receipts,
+    transaction,
+    transaction_index,
+    is_pending
+):
     receipt = receipts[transaction_index]
 
     if transaction.to == b'':
@@ -105,6 +111,7 @@ def serialize_transaction_receipt(block, receipts, transaction, transaction_inde
             serialize_log(block, transaction, transaction_index, log, log_index, is_pending)
             for log_index, log in enumerate(receipt.logs)
         ],
+        'state_root': receipt.state_root,
     }
 
 
--- a/eth_tester/main.py
+++ b/eth_tester/main.py
@@ -8,6 +8,7 @@ import functools
 
 from cytoolz import (
     dissoc,
+    assoc
 )
 from cytoolz.itertoolz import (
     remove,
@@ -24,6 +25,7 @@ from eth_utils import (
     to_dict,
     to_list,
     to_tuple,
+    to_int
 )
 
 from eth_tester.backends import (
@@ -336,6 +338,20 @@ class EthereumTester(object):
         raw_receipt = self.backend.get_transaction_receipt(raw_transaction_hash)
         self.validator.validate_outbound_receipt(raw_receipt)
         receipt = self.normalizer.normalize_outbound_receipt(raw_receipt)
+
+        if 'FORK_BYZANTIUM' in self.get_supported_forks():
+            byzantium_fork_block = self.get_fork_block('FORK_BYZANTIUM')
+            transaction = self.get_transaction_by_hash(transaction_hash)
+        else:
+            byzantium_fork_block = None
+        if((byzantium_fork_block is not None) and
+           (transaction['block_number'] is not None) and
+           (transaction['block_number'] >= byzantium_fork_block)
+           ):
+            status = to_int(receipt.pop('state_root'))
+            if status > 1:
+                raise ValidationError('Invalid status value: only 0 or 1 are valid')
+            return assoc(receipt, 'status', status)
         return receipt
 
     #
--- a/eth_tester/backends/mock/serializers.py
+++ b/eth_tester/backends/mock/serializers.py
@@ -51,4 +51,5 @@ def serialize_receipt(transaction, block, transaction_index, is_pending):
         partial(assoc, key='block_number', value=block_number),
         partial(assoc, key='block_hash', value=block_hash),
         partial(assoc, key='transaction_index', value=transaction_index),
+        partial(assoc, key='state_root', value=b'\x00'),
     )
--- a/eth_tester/backends/pyethereum/serializers.py
+++ b/eth_tester/backends/pyethereum/serializers.py
@@ -51,6 +51,7 @@ def serialize_transaction_receipt(block,
             serialize_log(block, transaction, transaction_index, log, log_index, is_pending)
             for log_index, log in enumerate(transaction_receipt.logs)
         ],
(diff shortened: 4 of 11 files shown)


eth_tester/main.py
```diff
@@ -333,9 +333,13 @@ def get_transaction_receipt(self, transaction_hash):
         raw_transaction_hash = self.normalizer.normalize_inbound_transaction_hash(
             transaction_hash,
         )
+        transaction = self.get_transaction_by_hash(transaction_hash)
         raw_receipt = self.backend.get_transaction_receipt(raw_transaction_hash)
         self.validator.validate_outbound_receipt(raw_receipt)
         receipt = self.normalizer.normalize_outbound_receipt(raw_receipt)
+        byzantium_fork_block = self.get_fork_block('FORK_BYZANTIUM')
+        if byzantium_fork_block is None or byzantium_fork_block > transaction['block_number']:
+            del receipt['status']
```
pipermerriam: Instead of removing the `status` field here, what about changing this so that it instead checks to see if we are in the byzantium vm and if `True`, removes the `state_root` field and adds the `status` field.  That would remove any need to change implementation in individual backends.

pcppcp: the main problem I've run into are validators - there is no easy way of having them check the status conditionally and the status must be set by the serializers of individual backends.

pipermerriam: So my thought is that:

1. All of the backends would **not** set status.
2. Any validators/normalizers that apply at the backend level get stripped of any knowledge of the `status` field
3. All backends include `state_root` (which for byzantium and later will always bee `0x0` or `0x1`.
4. At the `EthereumTester` level, *if* we are in the byzantium rules, we pop the `state_root` off of the return value, and inject a new `status` field computed from the `state_root` field.

pcppcp: Please check the diff now, I've updated it according to your comments.

eth_tester/main.py
```diff
@@ -333,9 +333,17 @@ def get_transaction_receipt(self, transaction_hash):
         raw_transaction_hash = self.normalizer.normalize_inbound_transaction_hash(
             transaction_hash,
         )
+        transaction = self.get_transaction_by_hash(transaction_hash)
         raw_receipt = self.backend.get_transaction_receipt(raw_transaction_hash)
         self.validator.validate_outbound_receipt(raw_receipt)
         receipt = self.normalizer.normalize_outbound_receipt(raw_receipt)
+
+        byzantium_fork_block = None
```
pipermerriam: Instead of this can you place this in an `else` block for the following `if` statement.

eth_tester/main.py
```diff
@@ -333,9 +333,17 @@ def get_transaction_receipt(self, transaction_hash):
...
+        if 'FORK_BYZANTIUM' in self.get_supported_forks():
+            byzantium_fork_block = self.get_fork_block('FORK_BYZANTIUM')
+        status = int.from_bytes(receipt.pop('state_root'), byteorder='big')
```
pipermerriam: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt

According to the spec, we should only remove the `state_root` *if* we are putting the `status` field in, so this should probably be within the next `if` block.

eth_tester/main.py
```diff
@@ -333,9 +333,17 @@ def get_transaction_receipt(self, transaction_hash):
...
+        if byzantium_fork_block is not None and byzantium_fork_block <= transaction['block_number']:
+            receipt['status'] = status
```
pipermerriam: Instead of mutating this value, we generally have chosen to use non-mutative approach like:

```
from cytoolz import assoc

return assoc(receipt, 'status', status)
```

This generates a new dictionary rather than mutating the existing one.

eth_tester/backends/pyevm/serializers.py
```diff
@@ -93,7 +99,7 @@ def serialize_transaction_receipt(block, receipts, transaction, transaction_inde
     else:
         origin_gas = receipt.gas_used - receipts[transaction_index - 1].gas_used
 
-    return {
+    ret = {
```
carver: I don't think this ^ change is necessary, so you could revert it. But if you want to keep it, the variable would be easier to understand with a descriptive name -- something like `serialized_receipt`.

eth_tester/main.py
```diff
@@ -333,9 +334,18 @@ def get_transaction_receipt(self, transaction_hash):
...
+        transaction = self.get_transaction_by_hash(transaction_hash)
         raw_receipt = self.backend.get_transaction_receipt(raw_transaction_hash)
         self.validator.validate_outbound_receipt(raw_receipt)
         receipt = self.normalizer.normalize_outbound_receipt(raw_receipt)
+
+        if 'FORK_BYZANTIUM' in self.get_supported_forks():
+            byzantium_fork_block = self.get_fork_block('FORK_BYZANTIUM')
+        else:
+            byzantium_fork_block = None
+        if byzantium_fork_block is not None and byzantium_fork_block <= transaction['block_number']:
```
carver: The second comparison reads like a [yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions) to me. The order reversed, I prefer.

eth_tester/main.py
```diff
@@ -333,9 +334,18 @@ def get_transaction_receipt(self, transaction_hash):
...
+            status = int.from_bytes(receipt.pop('state_root'), byteorder='big')
```
carver: Another option here is to use our conversion utility here:

```py
from eth_utils import to_int

to_int(receipt.pop('state_root'))
```

(and fail validation if `status > 1` since that is not a valid response according to [EIP 658](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md#specification))

eth_tester/backends/pyevm/serializers.py
pcppcp: yeah just `return` is sufficient, I had other code manipulating the dict contents there and then I forgot to revert it back.

pcppcp:
Hey guys! Do you have any estimate when can we expect this to be merged & published on PyPi? It would really help us move forward with Raiden developement.
Thanks!

carver:
Ah, I thought this was in progress because of the broken test. Now I see that the failure is due to #103 -- I'll merge now. I'd like to get 103 solved before the next release (as well as #104). I hope to get started on those today.

carver:
Will likely be able to release in the next 24-48 hrs @pcppcp

pcppcp:
Awesome, thanks a lot!

carver:
@pcppcp beta 27 released! Please post an issue or ping me in https://gitter.im/ethereum/eth-tester if you run into any issues.

PR #102, opened 2018-06-07
Fix gas_used calculation in py-evm transaction receipt

NIC619:
### What was wrong?
`gas_used` of the transaction should be
```python
block_receipts[transaction_index].gas_used - block_receipts[transaction_index - 1].gas_used
```
### How was it fixed?
set `origin_gas` as `receipts[transaction_index - 1].gas_used`

#### Cute Animal Picture

Cute animal picture


carver:
The pyethereum test failure is not related to the PR. Captured in #103

carver:
Would you be up for writing a simple/short test to make sure this doesn't regress? Obviously we didn't have any coverage here.

NIC619:
@carver Sorry for the late reply. Added a test to check that `cumulative_gas_used` and `gas_used` are computed correctly.
Unfortunately the same goes for pyethereum and I'm working on the same patch and test for pyethereum right now. Will push it later(or tomorrow morning)

carver:
Ok, thanks @NIC619 !

If you rebase on master, the build failure in pyethereum16 should be resolved.

tox.ini
```diff
@@ -14,7 +14,7 @@ commands=
     core: py.test {posargs:tests/core}
     pyethereum16: py.test {posargs:tests/backends/test_pyethereum16.py}
     pyethereum21: py.test {posargs:tests/backends/test_pyethereum21.py}
-    pyevm: py.test {posargs:tests/backends/test_pyevm.py}
+    pyevm: py.test {posargs:tests/backends/test_pyevm.py tests/backends/test_pyevm_serializer.py}
```
carver: If you put the test into `https://github.com/ethereum/eth-tester/blob/master/eth_tester/utils/backend_testing.py`, then it will get run by all backends, and this line change can be dropped.

NIC619:
@carver Rebased and also fix the `gas_used` calculation in `MockBackend`. It's ready for merge now.

eth_tester/utils/backend_testing.py
```diff
@@ -1208,6 +1208,31 @@ def test_delete_filter(self, eth_tester):
...
+    #
+    def test_receipt_gas_used_computation(self, eth_tester):
+        eth_tester.disable_auto_mine_transactions()
+
+        tx_hashes = []
+        for i in range(4):
+            tx = {
+                'from': eth_tester.get_accounts()[i],
+                'to': eth_tester.get_accounts()[i + 1],
+                'gas': (i + 1) * 20000 + 10000,
```
carver: The fact that this is calculated makes it seem like the gas amount is relevant to the test, but it isn't AFAICT. This would be just as effective as a constant, like `30000`, right?

carver:
Since I didn't hear back today, and am going offline for a few days, I'll go ahead and merge.

PR #110, opened 2018-06-21
Naive suggested fix for #109

### LINKED ISSUE #109: Calling Filter.remove(*values) with no values produces unhandled error
* Version: 0.1.0b27
* Python: 3.6.5
* OS: linux

### What was wrong?
I am using the snapshot functionality to catch and revert a TransactionFailed error when automining transactions. I often (70% of the time in my test suite) run into an issue when reverting to a snapshot where it catches the following unhandled exception in this function call:
https://github.com/ethereum/eth-tester/blob/c51beb505f55f3dfd687d8b092a2afbb28452a0f/eth_tester/utils/filters.py#L47-L60

```python
    self._t.revert_to_snapshot(self._snapshot_id)
eth_tester/main.py:543: in revert_to_snapshot
    self._revert_log_filter(log_filter)
eth_tester/main.py:590: in _revert_log_filter
    filter.remove(*values_to_remove)
eth_tester/utils/filters.py:62: in remove
    in self.get_all()

---------------------------------------------------------------------------------------------

.0 = <tuple_iterator object at 0x7f2585dfdef0>

    value
    for value
    in self.get_all()
>   if value not in values_to_remove
            ]
E           TypeError: unhashable type: 'dict'

eth_tester/utils/filters.py:63: TypeError

```

When this error occurs, the following variables of interest are set:
```python
values = ()  # Empty tuple

# Tuple of 1 dict
self.get_all() = ({'type': 'mined', 'log_index': 0, 'transaction_index': 0, 'transaction_hash': b'_\x0f;\xd1n/@pq\xcaL\xea\xd1\x11\xe5\x86\x03\xc3\xe9.\xd4\x88\x8b9\xfaHD\xc3\xa3;V\x1d', 'block_hash': b'\xcf\x9b\xec\xcaD[\x7fm\xc0KJ\x1c{\x80.Z\t\xa8\xb7fV\xe5\xb9X\x05\xe0N\x85E\xb0\xc0.', 'block_number': 2, 'address': b'\xf2\xe2F\xbbv\xdf\x87l\xef\x8b8\xae\x84\x13\x0fOU\xde9[', 'data': b'', 'topics': (b'\x8b\xe0\x07\x9cS\x16Y\x14\x13D\xcd\x1f\xd0\xa4\xf2\x84\x19I\x7f\x97"\xa3\xda\xaf\xe3\xb4\x18okdW\xe0', b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00+Z\xd5\xc4y\\\x02e\x14\xf81|z!^!\x8d\xcc\xd6\xcf', b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00+Z\xd5\xc4y\\\x02e\x14\xf81|z!^!\x8d\xcc\xd6\xcf')},)

values_to_remove = set()  # Empty set

```

### How can it be fixed?
I haven't been able to replicate this issue as it is inconsistent for me during runtime in my testing tool, but I think handling when `values` is unset is fairly obvious, since this basic Python also raises the same error:
```python
[v for v in ({},) if v not in set()]

TypeError: unhashable type: 'dict'
```

I can submit a quick fix for this, but I need a little more instruction on how to handle this scenario

I am thinking of suggesting the following:
```python
    queued_values = self.get_changes()
    if values_to_remove:
        self.values = [
            value
            for value
            in self.get_all()
            if value not in values_to_remove
        ]
    else:
        self.values = self.get_all()
```

fubuloubu:
### What was wrong?
Fixes #109 as discussed

### How was it fixed?
Avoid handling set check with a dict

#### Cute Animal Picture

Pug


fubuloubu:
might warrant an additional test case for this behavior

pipermerriam:
This doesn't seem right as it doesn't fix the fact that `get_all()` is returning dictionaries and we cannot do `in` checks with a `dict` as the value.

pipermerriam:
Nevermind, this is the right fix (was mis-understanding the code).  👍

eth_tester/utils/filters.py
```diff
@@ -52,16 +52,21 @@ def remove(self, *values):
             values_to_remove = values
 
         queued_values = self.get_changes()
```
pipermerriam: This should be moved to inside of the `if` block. and the `else` block at the bottom should be dropped.

fubuloubu: Basically, abort doing anything if `values_to_remove` is empty?

```python
    def remove(self, *values):
        try:
            values_to_remove = set(values)
        except TypeError:
            # log filters are dicts which are not hashable
            values_to_remove = values

        if values_to_remove:
            queued_values = self.get_changes()
            self.values = [
                value
                for value
                in self.get_all()
                if value not in values_to_remove
            ]
            for value in queued_values:
                if value in values_to_remove:
                    continue
                self.queue.put_nowait(value)
```

fubuloubu:
Awesome, this was my largest problem I couldn't figure out with pytest-ethereum. I'll have a full test case suite working for one of my projects when this PR is accepted!

eth_tester/utils/filters.py
fubuloubu: Alternative suggestion:
```python
    def remove(self, *values):
        if not values:
            # Nothing to do!
            return

        try:
            values_to_remove = set(values)
        except TypeError:
            # log filters are dicts which are not hashable
            values_to_remove = values

        queued_values = self.get_changes()
        self.values = [
            value
            for value
            in self.get_all()
            if value not in values_to_remove
        ]
        for value in queued_values:
            if value in values_to_remove:
                continue
            self.queue.put_nowait(value)
```

pipermerriam: Later suggestion is even better.

fubuloubu:
Made the requested fix and rolled back old commits

Issue #112, opened 2018-06-22
Creating a filter generates duplicate events in other filters


pcppcp:
* Version: 0.1.0b27
* Python: 3.5
* OS: linux

### What was wrong?

When creating a new filter using `create_log_filter()`, `_process_blocks_logs()` is called to populate filter with past events.
This method however calls `add()` method for all existing filters:
```python
    def _process_block_logs(self, block):
        for _, filter in self._log_filters.items():
            for transaction_hash in block['transactions']:
                receipt = self.get_transaction_receipt(transaction_hash)
                for log_entry in receipt['logs']:
                    raw_log_entry = self.normalizer.normalize_inbound_log_entry(log_entry)
                    filter.add(raw_log_entry)
```

### How can it be fixed?

`Filter::add()` method should check if the log entry already exists in `Filter::values` before appending/processing it.

pcppcp:
Might be also related to this `web3` issue

https://github.com/ethereum/web3.py/issues/907

carver:
Closed by #113

PR #113, opened 2018-06-22
Fix duplicate log when using multiple filters

pcppcp:
### What was wrong?
Issue #112 

### How was it fixed?
`Filter::add()` wont process an item if log with a same txhash already exists.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/main.py
+++ b/eth_tester/main.py
@@ -415,10 +415,16 @@ class EthereumTester(object):
     #
     def _process_block_logs(self, block):
         for _, filter in self._log_filters.items():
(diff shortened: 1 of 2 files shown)


carver:
Thanks for hunting this down @pcppcp ! 

Just added a small tweak so that we only process the logs for the new filter. That way we don't have to track any "already seen" values.

carver:
I'm cutting a release today, so I'll go ahead and merge this. Please continue the conversation here if there's anything more you want to contribute, and/or if you see problems with the commit I added.

PR #115, opened 2018-07-19
Revert reason message

jacqueswww:
### What was wrong?

Fixes part of https://github.com/ethereum/py-evm/issues/977.

Adds decoding of EIP838 strings to TransactionFailed exception.

### How was it fixed?

Extended the `throws_contract.py` utility to contain 2 contracts, one for throwing and one for reverting.


eth_tester/backends/pyevm/main.py
```diff
@@ -507,6 +511,16 @@ def call(self, transaction, block_number="latest"):
...
-            raise TransactionFailed(str(computation._error))
+            msg = str(computation._error)
+
+            # Check to see if it's a EIP838 standard error, with ABI signature
+            # of Error(string). If so - extract the message/reason.
+            if isinstance(computation._error, EVMRevert) and \
+               len(computation._error.args) > 0 and \
+               computation._error.args[0][:4] == b'\x08\xc3y\xa0':
+                error_str = computation._error.args[0][36:]
+                msg = decode_single('bytes', error_str)
```
pipermerriam: I believe this can throw an `eth_abi.DecodingError` which should be caught and handled.

eth_tester/backends/pyevm/main.py
```diff
@@ -507,6 +511,16 @@ def call(self, transaction, block_number="latest"):
...
         )
         if computation.is_error:
-            raise TransactionFailed(str(computation._error))
+            msg = str(computation._error)
+
+            # Check to see if it's a EIP838 standard error, with ABI signature
+            # of Error(string). If so - extract the message/reason.
+            if isinstance(computation._error, EVMRevert) and \
+               len(computation._error.args) > 0 and \
+               computation._error.args[0][:4] == b'\x08\xc3y\xa0':
```
pipermerriam: Can you move this string literal to a named constant at the top level of this module.

eth_tester/backends/pyevm/main.py
```diff
@@ -507,6 +511,16 @@ def call(self, transaction, block_number="latest"):
             block_number,
         )
         if computation.is_error:
-            raise TransactionFailed(str(computation._error))
+            msg = str(computation._error)
+
+            # Check to see if it's a EIP838 standard error, with ABI signature
+            # of Error(string). If so - extract the message/reason.
+            if isinstance(computation._error, EVMRevert) and \
```
pipermerriam: We've largely avoided this style of line continuation.  My suggestion would be to move this to a small utility function.

```python
def is_eip838_error(error):
    if not isinstance(err, EVMRevert):
        return False
    elif ...:
        ...
```

This will allow you to add unit tests for this specific logic which I'd like to see added.

Then in this function body you can do

```python
if is_eip838_error(computation._error):
    ...
```

eth_tester/utils/throws_contract.py
```diff
@@ -29,46 +40,78 @@
     "080fd5b606060a8565b005b3415606c57600080fd5b607260b6565b60405180828152602001915050"
     "60405180910390f35b3415609257600080fd5b60a6600480803590602001909190505060bc565b005"
     "b6000151560b457600080fd5b565b60005481565b80600081905550505600a165627a7a72305820d5"
-    "8a6b595eeb1a765e71924eec6e96c98c1fe9e90876ec232593b9ebb9c686500029"
+    "8a6b595eeb1a765e71924eec6e96c98c1fe9e90876ec232593b9ebb9c686500029",
```
pipermerriam: I'd rather put the new bytecode in a new variable.  Something like `REVERT_BYTECODE`

eth_tester/utils/throws_contract.py
```diff
@@ -29,46 +40,78 @@
...
+)
 
 
-def _deploy_throws(eth_tester):
+def _deploy_throws(eth_tester, contract_idx=0):
     deploy_hash = eth_tester.send_transaction({
         "from": eth_tester.get_accounts()[0],
         "gas": 500000,
-        "data": THROWS_BYTECODE,
+        "data": THROWS_BYTECODE[contract_idx],
```
pipermerriam: Can we drop the `contract_idx` in favor of distinct contracts?

jacqueswww: Sure, I was initially going to create own `revert_contract.py` but then realised the _throw function would be exact replicas - which lead me to making multiple contracts.

What other type of identifier could I use on the function call level, assuming I keep the _*throws functions (dict instead of tuple with contract_name arg)?

eth_tester/backends/pyevm/main.py
```diff
@@ -507,6 +523,19 @@ def call(self, transaction, block_number="latest"):
...
         if computation.is_error:
-            raise TransactionFailed(str(computation._error))
+            msg = str(computation._error)
+
+            # Check to see if it's a EIP838 standard error, with ABI signature
+            # of Error(string). If so - extract the message/reason.
+            if self.is_eip838_error(computation._error):
+                error_str = computation._error.args[0][36:]
+                try:
+                    msg = decode_single('bytes', error_str)
```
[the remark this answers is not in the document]
jacqueswww: Made changes and overrode previous commit ;)

eth_tester/backends/pyevm/main.py
```diff
@@ -490,6 +497,17 @@ def estimate_gas(self, transaction):
...
         return self.chain.estimate_gas(spoofed_transaction)
 
+    def is_eip838_error(self, error):
+        if not isinstance(error, EVMRevert):
+            return False
+        elif len(error.args) == 0:
+            return False
+
+        try:
+            return error.args[0][:4] == EIP838_SIG
```
pipermerriam: I suspect this can also throw an `IndexError` were the exception to have no arguments.  Probably worth handling.

pipermerriam:
merging without pyethereum backend passing.  Those are slated to be deprecated and now that we've formally deprecated pyethapp, we should be very close to being able to drop pyethereum.

PR #117, opened 2018-07-24
Compatibility with py-evm v0.2.0-alpha.28

carver:
- switch from `Chain` to `MiningChain` which means we have to stop using the prebuilt tester chain (although py-evm should probably switch to using a `MiningChain`)
- change module name from `evm` to `eth`
- change a keyword argument from header to `at_header`
- drop an unused `configure_forks` feature that was broken anyway, last I checked.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -37,7 +37,6 @@ from eth_tester.constants import (
 from eth_tester.exceptions import (
     BlockNotFound,
     TransactionNotFound,
-    UnknownFork,
     TransactionFailed,
     BackendDistributionNotFound,
 )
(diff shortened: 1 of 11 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -339,8 +343,7 @@ def get_fork_block(self, fork_name):
             raise UnknownFork("Unknown fork name: {0}".format(fork_name))
 
     def configure_forks(self):
-        args, kwargs = _mk_fork_configuration_params(self.fork_config)
-        self.chain.configure_forks(*args, **kwargs)
+        pass
```
pipermerriam: Maybe this should raise an exception?  Silently passing seems weird.

carver: Thoughts on removing all the runtime fork configuration options? (`set_fork_block`, `get_supported_forks`, `get_fork_block`)

I'm not sure they work, and even if they do right now, I'm not sure it's a good idea to enable. That seems like the kind of thing that should be configured at setup time instead of runtime. That's the way py-evm would work most safely.

carver: So that comes with another quirk: Pyethereum 1.6 doesn't support Byzantium rules. I'm not keen to work hard to support a deprecated backend, so the current solution is to drop pyethereum1.6 support. Thoughts?

pipermerriam: Dropping all of pyethereum support is on my 1.0 tick-list so yes.

PR #118, opened 2018-07-31
Fix topic_list parameter implementation (matching and validation) to align with JSON-RPC spec

### LINKED ISSUE #119: nested topic arrays should be interpreted as match options for the topic at the nested array position
* Version:0.1.0b29
* Python: 3.6

### What was wrong?
See review comment here: https://github.com/ethereum/web3.py/pull/976/files/bec2e4e082c2ac58fd38063dec5f5a22641bc3cc

To summarize: we were interpreting the nested arrays in topic lists as full variants of the topic list, rather than options for the topic at the nested array position.

### How can it be fixed?
Several methods, validators etc. need to be updated to follow the spec here:

https://github.com/ethereum/wiki/wiki/JSON-RPC#a-note-on-specifying-topic-filters

dylanjw:
### What was wrong?

Validators disallow topics lists such as:
[TOPICA, [TOPICB, TOPICC], TOPICD]

EDIT: fixes #119

### How was it fixed?
The topic validators were extended to check the validity of single items along with flat topic arrays.

#### Cute Animal Picture

image

### DIFF
--- a/tests/core/filter-utils/test_filter_helpers.py
+++ b/tests/core/filter-utils/test_filter_helpers.py
@@ -15,7 +15,7 @@ from eth_tester.utils.filters import (
     check_if_log_matches,
     is_topic,
     is_flat_topic_array,
-    is_nested_topic_array,
+    is_valid_with_nested_topic_array,
     is_topic_array,
 )
 
@@ -131,17 +131,17 @@ NESTED_TOPICS_E = (TOPIC_A, TOPICS_MANY, TOPICS_EMPTY)
         ([b'a', None, b'b'], False),
         (list(), False),
         ([None], False),
-        (TOPIC_A, False),
-        (TOPICS_EMPTY, False),
-        (TOPICS_SINGLE_NULL, False),
-        (TOPICS_MANY, False),
-        (TOPICS_MANY_WITH_NULL, False),
         (([],), False),
         (([tuple()],), False),
         ([tuple()], False),
         ((tuple(), []), False),
         ((TOPICS_EMPTY, (b'arst',)), False),
+        (TOPIC_A, False),
+        (TOPICS_EMPTY, False),
         # good values
+        (TOPICS_SINGLE_NULL, True),
+        (TOPICS_MANY, True),
+        (TOPICS_MANY_WITH_NULL, True),
         (NESTED_TOPICS_A, True),
         (NESTED_TOPICS_B, True),
         (NESTED_TOPICS_C, True),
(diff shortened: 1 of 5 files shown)


tests/core/filter-utils/test_filter_helpers.py
```diff
@@ -308,38 +311,38 @@ def test_check_if_to_block_match(block_number, _type, to_block, expected):
...
-        (TOPICS_A_A, FILTER_MATCH_ONLY_A, True),
-        (TOPICS_A_B, FILTER_MATCH_ONLY_A, True),
-        (TOPICS_A_C, FILTER_MATCH_ONLY_A, True),
-        (TOPICS_A_B_C, FILTER_MATCH_ONLY_A, True),
-        (TOPICS_A_C_B, FILTER_MATCH_ONLY_A, True),
-        (TOPICS_B_A, FILTER_MATCH_ONLY_A, False),
-        (TOPICS_B_C, FILTER_MATCH_ONLY_A, False),
-        (TOPICS_B_A_C, FILTER_MATCH_ONLY_A, False),
-        (TOPICS_B_C_A, FILTER_MATCH_ONLY_A, False),
+        (TOPICS_EMPTY, FILTER_MATCH_ONE_OR_MORE, False),
```
dylanjw: This one im not sure on.  Intuitively I would think `(None,)` should match all non-anonymous events, but that is assuming.

tests/core/filter-utils/test_filter_helpers.py
```diff
@@ -308,38 +311,38 @@ def test_check_if_to_block_match(block_number, _type, to_block, expected):
...
+        (TOPICS_B_A_C, FILTER_MATCH_A, False),
+        (TOPICS_B_C_A, FILTER_MATCH_A, False),
         # length 2 matches
-        (TOPICS_EMPTY, FILTER_MATCH_ANY_TWO, False),
-        (TOPICS_A_A, FILTER_MATCH_ANY_TWO, True),
-        (TOPICS_A_B, FILTER_MATCH_ANY_TWO, True),
-        (TOPICS_ONLY_A, FILTER_MATCH_ANY_TWO, False),
-        (TOPICS_ONLY_B, FILTER_MATCH_ANY_TWO, False),
-        (TOPICS_ONLY_C, FILTER_MATCH_ANY_TWO, False),
+        (TOPICS_EMPTY, FILTER_MATCH_TWO_OR_MORE, False),
```
dylanjw: Same assumption here.  That a None value matches any topic that exists at that position, but not an absent topic.

tests/core/filter-utils/test_filter_helpers.py
carver: The parity version makes more sense to me. You can already get the geth like functionally by removing the None.

dylanjw: Parity doesn't allow things like matching any one, or any two, like go-ethereum does.  go-ethereum essentially treats None as a place specific wildcard.  On the other hand None feels like a bad choice for a wildcard symbol, and is more intuitive that it provide no actual matching function.

carver: Ah, right, I got them switched up. The geth one makes more sense to me, too. :+1:

PR #121, opened 2018-08-16
fixed filters to work correctly when to_block points to latest block

voith:
### What was wrong?

filter didn't include logs  from block with `blockNumber=to_block`, when `to_block` was passed as a parameter to `create_log_filter`.

### How was it fixed?
Increased `upper_bound` by 1 when `to_block` is passed. 

https://github.com/ethereum/eth-tester/blob/a0ac1bc9dd5c467e41e6d841a374b28b72fa8701/eth_tester/main.py#L611
#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/utils/backend_testing.py
+++ b/eth_tester/utils/backend_testing.py
@@ -1216,6 +1216,28 @@ class BaseTestBackendDirect(object):
         direct_logs_all = eth_tester.get_logs(from_block=0)
         assert len(logs_changes) == len(logs_all) == len(direct_logs_all) == 2
 
+    def test_log_filter_includes_latest_block_with_to_block(self, eth_tester):
+        self.skip_if_no_evm_execution()
+
+        emitter_address = _deploy_emitter(eth_tester)
+        no_of_events = 2
+        for i in range(1, no_of_events + 1):
+            _call_emitter(
+                eth_tester,
+                emitter_address,
+                'logSingle',
+                [EMITTER_ENUM['LogSingleWithIndex'], i],
+            )
+
+        filter_any_id = eth_tester.create_log_filter(
+            from_block=0,
+            to_block=eth_tester.get_block_by_number('latest')['number']
+        )
+
+        logs_changes = eth_tester.get_only_filter_changes(filter_any_id)
+        logs_all = eth_tester.get_all_filter_logs(filter_any_id)
+        assert len(logs_changes) == len(logs_all) == no_of_events
+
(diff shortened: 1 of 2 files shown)


voith:
Here's a failing [build](https://circleci.com/gh/ethereum/eth-tester/1338?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

voith:
The `pyethereum` tests are failing but the `py-evm` tests are passing. This now seems like a backend specific issue.

voith:
Will take a look at this later today. I feel that there is an issue with the latest block returned.

voith:
using `latest` instead of `pending` fixed the problem. Looking at the code again now my second commit was wrong as the `upper_bound` was calculated using the `pending` block which is already +1.

However, looks like the real problem is with the pending block returned between backends. I am not yet sure though. I need to dig in.

eth_tester/utils/backend_testing.py
```diff
@@ -1203,17 +1203,24 @@ def test_log_filter_includes_old_logs(self, eth_tester):
...
+            to_block=eth_tester.get_block_by_number('latest')['number']
+        )
+
         logs_changes = eth_tester.get_only_filter_changes(filter_any_id)
         logs_all = eth_tester.get_all_filter_logs(filter_any_id)
-        direct_logs_all = eth_tester.get_logs(from_block=0)
+        direct_logs_all = eth_tester.get_logs(
+            from_block=0,
+            to_block=eth_tester.get_block_by_number('latest')['number']
+        )
```
carver: I think we should leave in an old test without setting `to_block`, and add a new one that sets `to_block`.

carver:
You've been neck deep in filters, lately @dylanjw - thoughts?

eth_tester/utils/backend_testing.py
voith: sure wll do

voith:
I rushed to the solution earlier as I needed an immediate fix. I have fixed this using a different approach.
There was no need to change `pending` -> `latest`. I was fixing this at the wrong place. 
`pending` is already `latest` + 1, but  `upper_bound = to_block` needed to be 
`upper_bound = to_block + 1` to get the range below working:
https://github.com/ethereum/eth-tester/blob/a0ac1bc9dd5c467e41e6d841a374b28b72fa8701/eth_tester/main.py#L611
cc @carver @dylanjw

carver:
LGTM :+1: as long as it makes sense to @dylanjw

PR #127, opened 2018-09-16
Made eth-abi a primary dependency

voith:
### What was wrong?
`eth-tester` would give the following error:

```
In [1]: from eth_tester import EthereumTester
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-72c9d4f19b88> in <module>()
----> 1 from eth_tester import EthereumTester

~/Projects/eth-tester/eth_tester/__init__.py in <module>()
      2 import sys
      3
----> 4 from .main import (  # noqa: F401
      5     EthereumTester,
      6 )

~/Projects/eth-tester/eth_tester/main.py in <module>()
     29 )
     30
---> 31 from eth_tester.backends import (
     32     get_chain_backend,
     33 )

~/Projects/eth-tester/eth_tester/backends/__init__.py in <module>()
     10     MockBackend,
     11 )
---> 12 from .pyevm import (  # noqa: F401
     13     PyEVMBackend,
     14     is_pyevm_available,

~/Projects/eth-tester/eth_tester/backends/pyevm/__init__.py in <module>()
      4     is_pyevm_available,
      5 )
----> 6 from .main import (  # noqa: F401
      7     PyEVMBackend,
      8 )

~/Projects/eth-tester/eth_tester/backends/pyevm/main.py in <module>()
     10 import rlp
     11
---> 12 from eth_abi import (
     13     decode_single
     14 )

ModuleNotFoundError: No module named 'eth_abi'
```
This was missed in #115 (4073ee2dd594bfa57b155c588c2492b02c6ccb11).
This wasn't detected by tests because `eth-abi` was testing dependency

### How was it fixed?
`eth-abi` was removed as a testing dependency and was made a primary dependency.

####  Bonus changes
-  `pycryptodome` or `pysha3` are optional dependencies of `py-evm` that are needed for `eth-tester` to work. Install them based on the python implementation available. (Haven't tested it with pypy manually, I'm relying on tests to tell me).
-  Added python3.7 to the list of supported python versions.

PS: I have added all these in separate commits to keep unrelated changes isolated.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,6 @@ extras_require = {
     'test': [
         'pytest>=3.2.1,<4.0.0',
         'pytest-xdist>=1.22.2,<2',
-        'eth-abi>=1.0.0-beta.1,<2',
         'eth-hash[pycryptodome]>=0.1.4,<1.0.0',
     ],
     'dev': [
@@ -55,6 +54,7 @@ setup(
         "rlp>=0.6.0,<2.0.0",
         "semantic_version>=2.6.0,<3.0.0",
         "eth-keys>=0.2.0-beta.3,<0.3.0",
+        "eth-abi>=1.0.0-beta.1,<2",
     ],
     extras_require=extras_require,
     setup_requires=['setuptools-markdown'],
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,8 @@ extras_require = {
         # Pin py-evm to exact version, until it leaves alpha.
         # EVM is very high velocity and might change API at each alpha.
         "py-evm==0.2.0a32",
+        "eth-hash[pysha3]>=0.1.4,<1.0.0;implementation_name=='cpython'",
+        "eth-hash[pycryptodome]>=0.1.4,<1.0.0;implementation_name=='pypy'",
     ],
 }
 
--- a/setup.py
+++ b/setup.py
@@ -73,5 +73,6 @@ setup(
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
     ],
 )


carver:
Just to explicitly state the implicit change: this treats eth-tester like an end product rather than a library. (by forcing a choice of eth-hash backend, for example)

I think this is reasonable, just wanted to make sure everyone is one the same page.

voith:
@carver My reasoning is as follows:
I checked `setup.py` in `py-evm`. The [extra-eth](https://github.com/ethereum/py-evm/blob/1006923c7d2f4f6e6317e172757967c7a4b7bf02/setup.py#L26-L27) requirements in `py-evm` are:
```
'eth-extra': [
        "coincurve>=8.0.0,<9.0.0",
        "eth-hash[pysha3];implementation_name=='cpython'",
        "eth-hash[pycryptodome];implementation_name=='pypy'",
        "plyvel==1.0.5",
    ],
```
Although a user still has the freedom to explicitly install any `eth-hash` backend, installations using `eth-extra` will bind the backend with the python implementation.
The change I made here is no different.

carver:
That's a good point, the `eth-tester[py-evm]` extra should probably just be on `py-evm[eth-extra]` rather than on `py-evm` and duplicating all those packages here.

voith:
@carver I had thought about using `py-evm[eth-extra]`. But there were two extra dependencies,  `plyvel` and `coincurve` for which there are no tests at the moment.
Also `"eth-hash[pysha3];implementation_name=='cpython'` and ` "eth-hash[pycryptodome];implementation_name=='pypy'"` are not pinned to any range at the moment in `py-evm`, which worried me. 

Probably,  I should have fixed the `eth-hash` change in `py-evm`, but to have `plylevel` as dependency, it would need some tests.

PR #134, opened 2018-10-04
Upgrade py-evm to alpha 33

carver:
### What was wrong?

Pinned to old version of py-evm

### How was it fixed?

- Upgraded py-evm
- Also, requires more recent eth-utils (for `ValidationError` import)

#### Cute Animal Picture

Cute animal picture


carver:
Oops, local `tox` run didn't regenerate, so missed some bugs.

PR #137, opened 2018-10-13
Cleanup address utils

### LINKED ISSUE #86: Cleanup address and account utils
### What is wrong?

- There is a duplicate util in both `eth_tester.utils.account` and `eth_tester.utils.address` for generating contract addresses.
- The `private_key_to_address` function should be removed in favor of using `eth-keys` API

onyb:
### What was wrong?

- `eth_tester.utils.account.generate_contract_address` and `eth_tester.utils.address.generate_contract_address` were duplicates.

- `eth_tester.utils.account.private_key_to_address` was not using the standard `eth_keys` API.

Closes #86.

### How was it fixed?

- removed the `eth_tester.utils.account.generate_contract_address`.
- modified `eth_tester.utils.address.generate_contract_address` to return a canonical address.
- **modified** the `eth_tester.utils.account.private_key_to_address` to use `eth_keys` API instead of the utils in the `sekp256k1` module.
- subsequently removed the `sekp256k1` and `jacobian` modules, which were not used anymore.

#### Cute Animal Picture

Cute animal picture


eth_tester/utils/address.py
```diff
@@ -2,8 +2,9 @@
 
 from eth_utils import (
     keccak,
+    to_canonical_address,
 )
 
 
 def generate_contract_address(address, nonce):
-    return keccak(rlp.encode([address, nonce]))[-20:]
+    return to_canonical_address(keccak(rlp.encode([address, nonce]))[-20:])
```
carver: What do you think about breaking this up into a few lines?

Something like:
```py
next_account_hash = keccak(rlp.encode([address, nonce]))
return to_canonical_address(next_account_hash[-20:])

onyb: Done, thanks.

PR #140, opened 2018-12-07
Add exception message to from

jacqueswww:
### What was wrong?

When supplying an invalid from address, one would receive a KeyError.

### How was it fixed?

Throws a more useful exception.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -473,7 +473,10 @@ class PyEVMBackend(object):
 
     def _get_normalized_and_signed_evm_transaction(self, transaction, block_number='latest'):
         if transaction['from'] not in self._key_lookup:
-            raise ValidationError('"from" key not available, does this account exist?')
+            raise ValidationError(
+                'No valid "from" key was provided in the transaction '
+                'which is required for transaction signing.'
+            )
         signing_key = self._key_lookup[transaction['from']]
         normalized_transaction = self._normalize_transaction(transaction, block_number)
         evm_transaction = self.chain.create_unsigned_transaction(**normalized_transaction)
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -39,7 +39,6 @@ from eth_tester.exceptions import (
     BlockNotFound,
     TransactionFailed,
     TransactionNotFound,
-    UnknownFork,
     ValidationError,
 )
 
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -35,10 +35,12 @@ from eth_tester.constants import (
     FORK_BYZANTIUM,
 )
 from eth_tester.exceptions import (
-    BlockNotFound,
-    TransactionNotFound,
-    TransactionFailed,
     BackendDistributionNotFound,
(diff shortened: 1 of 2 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -471,6 +472,8 @@ def _get_normalized_and_unsigned_evm_transaction(self, transaction, block_number
         return evm_transaction
 
     def _get_normalized_and_signed_evm_transaction(self, transaction, block_number='latest'):
+        if transaction['from'] not in self._key_lookup:
+            raise ValidationError('"from" key not available, does this account exist?')
```
pipermerriam: This message seems like it could be improved a bit.

> No `"from"` key was provided in the transaction which is required for transaction signing

A bit more concise and hints at how the user can fix it.

tests/backends/test_pyevm.py
```diff
@@ -41,88 +37,111 @@ def test_generate_custom_genesis_state(self):
...
-        genesis_state = PyEVMBackend._generate_genesis_state(overrides=state_overrides, num_accounts=3)
+        genesis_state = PyEVMBackend._generate_genesis_state(
+            overrides=state_overrides, num_accounts=3
+        )
         assert len(genesis_state) == 3
         for _public_address, account_state in genesis_state.items():
-            assert account_state['balance'] == state_overrides['balance']
-            assert account_state['code'] == b''
+            assert account_state["balance"] == state_overrides["balance"]
+            assert account_state["code"] == b""
```
pipermerriam: Not sure if this was an automated change but we tend to try and isolate these types of changes from real code changes.  No need to separate here but in future PR's try to split this up into a "cleanup" PR and a "fix things" PR

tests/backends/test_pyevm.py
```diff
@@ -41,88 +37,111 @@ def test_generate_custom_genesis_state(self):
...
+    def test_send_transaction_invalid_from(self, eth_tester):
+        accounts = eth_tester.get_accounts()
+        assert accounts, "No accounts available for transaction sending"
+
+        with pytest.raises(ValidationError) as excinfo:
+            self._send_and_check_transaction(
+                eth_tester, SIMPLE_TRANSACTION, ZERO_ADDRESS_HEX
+            )
+
+        assert "does this account exist?" in str(excinfo.value)
```
pipermerriam: You can remove this in favor of providing a `match` keyword to the `pytest.raises` call:

https://docs.pytest.org/en/latest/assert.html

eth_tester/backends/pyevm/main.py
jacqueswww: I am changing this message to `No valid "from" key was provided...` the reason is that one can specify a public key, but the the key isn't available to use (this is how I found the error, I specified a public key of a contract).

carver: If we wanted to get really fancy, we could list the valid keys and show the actual key, but that shouldn't hold up this PR.

tests/backends/test_pyevm.py
pipermerriam: Never a need to apologize for your code (though I'm sure you'll see me still do it plenty). :+1: 

Thanks for the contribution.

PR #160, opened 2019-06-05
Drop Python 3.5

fubuloubu:
### What was wrong?
All the cool kids were dropping Python 3.5. I wanted to be one of the cool kids.

### How was it fixed?
`grep`-foo

#### Cute Animal Picture
Kung Fu Panda


pipermerriam:
This is a potential place where `py-upgrade` could go through and replace things with `f-strings`.

pipermerriam:
wow, pyupgrade awesome.

PR #169, opened 2019-10-02
Default py-evm backend VM to Istanbul

carver:
### What was wrong?

eth-tester was defaulting to Constantinople (not even St Petersburg!)

### How was it fixed?

Default to Istanbul

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,11 @@
 Unreleased
 ----------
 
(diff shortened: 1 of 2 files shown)


carver:
This is released in v0.2.0-beta.3 
cc @palango

palango:
Thanks! That was fast!

PR #172, opened 2019-12-15
py-evm expects the block header, not the hash, as of v0.3.0-alpha.11.

jMyles:
### What was wrong?

Incompatibility with py-evm by dint of the backend.  Here's an example traceback: https://gist.github.com/jMyles/4a32dff65edf5b5b22ecbe4e24c24d70

The change to py-evm which caused this API modification is here: https://github.com/ethereum/py-evm/commit/bfdcf12fce8f2827ed802ec73bdb351ac99a5f8d

### How was it fixed?

By passing the full header instead of just the hash.

#### Cute Animal Picture
9aum5dcwzp441

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -304,7 +304,7 @@ class PyEVMBackend(BaseChainBackend):
         block = self.chain.get_block_by_hash(snapshot)
         chaindb = self.chain.chaindb
 
-        chaindb._set_as_canonical_chain_head(chaindb.db, block.header.hash, GENESIS_PARENT_HASH)
+        chaindb._set_as_canonical_chain_head(chaindb.db, block.header, GENESIS_PARENT_HASH)
         if block.number > 0:
             self.chain.import_block(block)
         else:
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ extras_require = {
     'py-evm': [
         # Pin py-evm to exact version, until it leaves alpha.
         # EVM is very high velocity and might change API at each alpha.
-        "py-evm==0.3.0a8",
+        "py-evm==0.3.0a11",
         "eth-hash[pysha3]>=0.1.4,<1.0.0;implementation_name=='cpython'",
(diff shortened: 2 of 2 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -304,7 +304,7 @@ def revert_to_snapshot(self, snapshot):
         block = self.chain.get_block_by_hash(snapshot)
         chaindb = self.chain.chaindb
 
-        chaindb._set_as_canonical_chain_head(chaindb.db, block.header.hash, GENESIS_PARENT_HASH)
+        chaindb._set_as_canonical_chain_head(chaindb.db, block.header, GENESIS_PARENT_HASH)
```
pipermerriam: wondering if we couldn't use the `eth.tools.builder` suite of tools to eliminate the use of private APIs here.  Instead of using this `snapshot` approach, we could just use the `eth.tools.builder.chain_split` API to split the chain each time we want mark a point for reversion... When we want to do the revert we would just use the forked chain and discard the main chain.

jMyles: Yeah, I can dig that.  Though it will take a bit of re-engineering, right?  Since `chain_sp[lit` wants to execute the callable itself, it will need to go... I'm not sure, somewhere down the stack.   Is `eth.tools.builder.copy` closer to a drop-in replacement?  Of course, we might actually *want* to do things more in the `chain_split` way than the "snapshot" way to begin with.

Anywho, are we good to go on a merge for this one for the moment while we work out the snapshot logic?

pipermerriam:
Probably get it reviewed > merged > released sometime Monday.

pipermerriam:
@cburgdorf could you cut a release of this tomorrow?

cburgdorf:
Sure, no problem.

cburgdorf:
@pipermerriam oh it looks like I don't have access rights for eth-tester on pypi yet.

cburgdorf:
This is now released as https://pypi.org/project/eth-tester/0.4.0b1/ notice that this does not yet include Muir Glacier so we'll likely follow up with another release soon after.

PR #175, opened 2020-01-09
Default py-evm backend VM to Muir Glacier

palango:
### What was wrong?

Istanbul fork was still used.

### How was it fixed?

Update to Muir Glacier. As nothing but the ice age delay changed this shouldn't cause trouble.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -147,17 +147,17 @@ def get_default_genesis_params(overrides=None):
 def setup_tester_chain(genesis_params=None, genesis_state=None, num_accounts=None):
     from eth.chains.base import MiningChain
     from eth.db import get_db_backend
-    from eth.vm.forks.istanbul import IstanbulVM
+    from eth.vm.forks.muir_glacier import MuirGlacierVM
 
-    class IstanbulNoProofVM(IstanbulVM):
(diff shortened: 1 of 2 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -147,17 +147,17 @@ def get_default_genesis_params(overrides=None):
 def setup_tester_chain(genesis_params=None, genesis_state=None, num_accounts=None):
     from eth.chains.base import MiningChain
     from eth.db import get_db_backend
-    from eth.vm.forks.istanbul import IstanbulVM
+    from eth.vm.forks.muir_glacier import MuirGlacierVM
 
-    class IstanbulNoProofVM(IstanbulVM):
-        """Istanbul VM rules, without validating any miner proof of work"""
+    class MuirGlacierNoProofVM(MuirGlacierVM):
```
cburgdorf: I was about to suggest a rename to something like `LastestMainetNoProofVM` but actually, this code will go away with the next version of Py-EVM anyway (which isn't released yet). The next version has an actual consensus engine API so subclassing to overwrite `validate_seal` won't be needed then anymore.

But anyway, that isn't released yet so that change looks good to me for now :+1:

Issue #154, opened 2019-04-17
estimateGas should work immediately after time_travel, without mining


berndbohmeier:
* Version: 0.1.0-beta.39
* Python: 3.6
* OS: linux

### What was wrong?
The documentation states 
```
Note: Time traveling will result in a new block being mined.
```

but this will actually not happen. The code just updates the header. 

### How can it be fixed?

Depending on what is the desired behavior, either the documentation or the code has to be fixed. 

Note: Because of this behavior, `TransactionFailed` is currently not working properly, as it will not be raised, even if the transaction will fail, because `estimateGas` will not take into account the changed time.

berndbohmeier:
I can open a PR on this, if I know which is the right behavior.

carver:
Without having dived into the details, I'd say that fewer side-effects is better. If you want a new block mined, then time travel and then mine the block.

cc @njgheorghita @kclowes to take a deeper look at preferred functionality.

berndbohmeier:
As I wrote in my note, the problem is then, that if you do not mine a block, you can not properly test your code currently. 

Something like this will not work:
```
tester.time_travel(time)
with pytest.raises(TransactionFailed):
   contract.functions.something_that_fails_because_of_timestamp().transact()
```

The transaction will fail, but `TransactionFailed` will not be raised, as mine_block will take into account the new timestamp, but `estimateGas` (which raised TransactionFailed) will not. 

A fix there could be then, that `estimateGas` should also take into account the new timestamp. 
I think it is wrong that you have to call `mine_block` to get the proper behavior.

carver:
> A fix there could be then, that `estimateGas` should also take into account the new timestamp.
> I think it is wrong that you have to call `mine_block` to get the proper behavior.

Agreed :+1:

Pet3ris:
FYI - the documentation still states this :).

carver:
Thanks for the reminder @Pet3ris -- I won't get a chance to dig into the functionality now, but at least we can make the docs be more accurate to the current state of the code. :)

pipermerriam:
I believe this was fixed in #178

PR #177, opened 2020-01-23
docs: time-travel note about needing to mine

carver:
### What was wrong?

#154 -- docs say that time traveling immediately results in a new block being mined. It does not.

### How was it fixed?

Updated the docs. (Later, it's the functionality that should change)

#### Cute Animal Picture

Cute animal picture


carver:
I'll rage-merge this doc-only change, and will fix up anything after, if @pipermerriam notices something.

PR #178, opened 2020-01-24
Mine blocks on time travel

njgheorghita:
### What was wrong?
#154 - if I'm not misunderstanding, all that needs to happen is to automine a single block with every `time_travel`. 

### How was it fixed?
Added `mine_blocks` functionality to `time_travel`

#### Cute Animal Picture
image

### DIFF
--- a/eth_tester/backends/mock/factory.py
+++ b/eth_tester/backends/mock/factory.py
@@ -357,7 +357,7 @@ def make_block_from_parent(parent_block, overrides=None):
     if 'timestamp' in overrides:
         yield 'timestamp', overrides['timestamp']
     else:
-        yield 'timestamp', parent_block['timestamp'] + 15
+        yield 'timestamp', parent_block['timestamp'] + 1
 
     if 'transactions' in overrides:
         yield 'transactions', overrides['transactions']
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -314,6 +314,7 @@ class PyEVMBackend(BaseChainBackend):
     # Meta
     #
     def time_travel(self, to_timestamp):
+        # timestamp adjusted by 1 b/c a second is added in mine_blocks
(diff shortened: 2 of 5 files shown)


eth_tester/backends/mock/factory.py
```diff
@@ -357,7 +357,7 @@ def make_block_from_parent(parent_block, overrides=None):
     if 'timestamp' in overrides:
         yield 'timestamp', overrides['timestamp']
     else:
-        yield 'timestamp', parent_block['timestamp'] + 15
```
njgheorghita: It seems like this `15` was arbitrary (probably just the avg block time) - as far as I can tell there are no other side affects to changing this.

eth_tester/backends/pyevm/main.py
```diff
@@ -314,7 +314,8 @@ def revert_to_snapshot(self, snapshot):
     # Meta
     #
     def time_travel(self, to_timestamp):
-        self.chain.header = self.chain.header.copy(timestamp=to_timestamp)
+        self.chain.header = self.chain.header.copy(timestamp=(to_timestamp - 1))
```
njgheorghita: Timestamps are automatically incremented by 1 sec / mined block [based on](https://github.com/ethereum/py-evm/blob/74a59be41b/eth/vm/base.py#L398)

eth_tester/backends/pyevm/main.py
```diff
@@ -314,7 +314,8 @@ def revert_to_snapshot(self, snapshot):
...
+        self.mine_blocks()
```
njgheorghita: The assumption here is that for every `time_travel()` we're only mining one block. Do we want to allow an optional arg to customize how many blocks are mined / time travel?

carver: Nah, if they want to mine more, they can just `mine_blocks()` again.

PR #180, opened 2020-02-24
Stop masking Revert error as TransactionFailed

kclowes:
### What was wrong?
Original issue is [here](https://github.com/ethereum/web3.py/issues/941).

I'm not sure what the original rationale was, but the `Revert` error was getting turned into a `TransactionFailed` error. I want to test that Revert is raised in web3.py in [this PR](https://github.com/ethereum/web3.py/pull/1585).

This is my first PR in eth-tester, so I'm not sure if there is something else that I need to do in py-evm, or what sorts of unintended consequences there will be in other libraries as this is a breaking change. 

### How was it fixed?
Removed `TransactionFailed` from the decorator that replaces exceptions. 

#### Cute Animal Picture

image

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -464,7 +464,7 @@ class PyEVMBackend(BaseChainBackend):
 
     @replace_exceptions({
         EVMInvalidInstruction: TransactionFailed,
-        EVMRevert: TransactionFailed})
+    })
     def estimate_gas(self, transaction):
         evm_transaction = self._get_normalized_and_unsigned_evm_transaction(assoc(
             transaction, 'gas', 21000))
--- a/eth_tester/utils/backend_testing.py
+++ b/eth_tester/utils/backend_testing.py
@@ -51,6 +51,11 @@ from .throws_contract import (
     _make_call_throws_transaction,
     _decode_throws_result,
 )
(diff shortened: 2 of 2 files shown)


kclowes:
On second thought, this is not the change I need. Will open another one!

carver:
> I'm not sure what the original rationale was, but the `Revert` error was getting turned into a `TransactionFailed` error. I want to test that Revert is raised in web3.py in [this PR](https://github.com/ethereum/web3.py/pull/1585).

FWIW, the rationale was to raise an exception that is not back-end dependent. `eth-tester` was trying to hide the pyethereum vs py-evm backends, so that the user can be agnostic. As py-evm has become more robust, this is maybe less important, and we can just make eth-tester like a convenience toolset with py-evm as the only explicit backend. Then we could pass through things like the underlying py-evm exceptions.

That's a chunk of work that's not done, and probably requires some discussion first.

kclowes:
Aha, I see. Thanks for the clarification! Is pyethereum still in use? It looks from the readme like it's not...

carver:
Nope. Really the only reason to keep the backend concept around is if we think it's important to support future backends (a world where py-evm becomes deprecated? a world where we want to support... some kind of c binding into a different evm engine?). I think the chances are small enough that I'm okay abandoning the backend idea. Though it would be nice to chat with Piper before actually digging into implementing that change.

PR #188, opened 2020-05-20
Make gas limit constant

### LINKED ISSUE #186: Gas limit is not constant
### What was wrong?
I try to set the gas_limit in the genesis block. This works, but if I now mine blocks the gas limit changes, where I would expect it to not change. 

```
genesis_parameters = get_default_genesis_params({"gas_limit": 8 * 10 ** 6})
chain = eth_tester.EthereumTester(
        eth_tester.PyEVMBackend(genesis_parameters=genesis_parameters)
    )
for _ in range(1000):
    chain.mine_block()
assert web3.eth.getBlock("latest").gasLimit == 8 * 10 ** 6
```

### How can it be fixed?
Copy the gas limit from the old block. I can prepare an PR for that, but first I want to check if you would accept that, as it could be that some are relying on the changing gas_limit.

berndbohmeier:
### What was wrong?
Fixes #186 
The gas limit changed because of py-evm backend

### How was it fixed?
By overriding the header creation function to just copy the gas limit from the parent. 

Let me know if this was the correct place to add the tests. 
I did not know why `validate_seal` was added as `@classmethod`, so maybe you want that for create_header_from_parent` too.
Also, this might be a breaking change for folks that rely on that behavior (one test here also expected that behavior). Another option would be to make it somehow configurable, but that would be more work. I think the default should be constant gas limit.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,7 +3,8 @@ Unreleased
 
 - Features
 
-	- Add support for gas estimate block identifiers
+  - Add support for gas estimate block identifiers
+  - Add support for custom virtual machine fork schedule in PyEVMBackend
 
 v0.4.0-beta.2
 -------------
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -144,24 +144,27 @@ def get_default_genesis_params(overrides=None):
     return genesis_params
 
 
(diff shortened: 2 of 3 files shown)


carver:
Oof, interesting, it force-closed this on the force-push with a typo, but then won't let me fix it. I can't reopen the PR or push the correct branch. I'll reopen in a sec, in a separate PR.

PR #189, opened 2020-05-21
support gas estimates with block identifier

marcgarreau:
### What was wrong?

Re: https://github.com/ethereum/web3.py/pull/1639, py-evm supports gas estimation with block identifiers, but eth-tester didn't yet pass those values through.

### How was it fixed?

`estimate_gas` now accepts and processes a `block_number`.

Feedback appreciated @carver (and anyone else interested); I processed the minimum surface area of this package required to get this far and get passing tests on the [Web3 side](https://github.com/ethereum/web3.py/pull/1639).

### To-Do:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,9 @@
 Unreleased
 -------------
 
-None
+- Features
+
+	- Add support for gas estimate block identifiers
 
 v0.4.0-beta.2
 -------------
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -465,12 +465,26 @@ class PyEVMBackend(BaseChainBackend):
     @replace_exceptions({
         EVMInvalidInstruction: TransactionFailed,
         EVMRevert: TransactionFailed})
-    def estimate_gas(self, transaction):
-        evm_transaction = self._get_normalized_and_unsigned_evm_transaction(assoc(
-            transaction, 'gas', 21000))
+    def estimate_gas(self, transaction, block_number="latest"):
+        evm_transaction = self._get_normalized_and_unsigned_evm_transaction(
+            assoc(transaction, 'gas', 21000),
+            block_number
+        )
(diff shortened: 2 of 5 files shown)


CHANGELOG
```diff
@@ -1,7 +1,9 @@
 Unreleased
 -------------
 
-None
+- Misc
+
+	- Add support for gas estimate block identifiers
```
carver, line 6: I'd call it a feature from eth-tester's perspective.

eth_tester/backends/pyevm/main.py
```diff
@@ -465,9 +465,9 @@ def _max_available_gas(self):
     @replace_exceptions({
         EVMInvalidInstruction: TransactionFailed,
         EVMRevert: TransactionFailed})
-    def estimate_gas(self, transaction):
+    def estimate_gas(self, transaction, block_number="latest"):
         evm_transaction = self._get_normalized_and_unsigned_evm_transaction(assoc(
-            transaction, 'gas', 21000))
+            transaction, 'gas', 21000), block_number)
```
carver, line 470: *nitpick*


I find this visually harder to parse, and [this style section](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#first-thing-on-its-own-line) implies that the preferred approach is:

```py
        evm_transaction = self._get_normalized_and_unsigned_evm_transaction(
            assoc(transaction, 'gas', 21000),
            block_number,
        )
```

It stands out to me more, now that there's >1 argument to the method.

pipermerriam:
Bike-shedding over the the way to test:

Contract with a toggle-able flag which just returns immediately or burns a bunch of gas in a loop.  Then you can set it and test on either side of the block number where the flag was toggled.  Agreed on this being lower priority for testing.  Like @carver said, probably better to document the testing approach in a "good first issue".

marcgarreau:
👍 new issue coming follow the merge.

Issue #106, opened 2018-06-12
py-evm backend completely ignores fork setup


carver:
### What was wrong?

As of alpha 26, there is no clean way to define the VM configuration at runtime. `MainnetTesterChain.configure_forks` is broken: it sets the configuration on the vm instance, and then `get_vm_class_for_block_number` ignores `vm_configuration` on the instance, using the class's `vm_configuration` instead.

### How can it be fixed?

No one noticed that fork choice stuff isn't working. Maybe just neuter the API to *officially* make it do nothing, and we can re-add the functionality when it's requested.

Also, probably remove `MainnetTesterChain.configure_forks` from py-evm since it's described as "footgun" and isn't working anyway.

pipermerriam:
![footgun-01](https://user-images.githubusercontent.com/824194/41302192-1c3489a6-6e27-11e8-8c3b-9c1fc597196c.jpg)

fubuloubu:
I was going to leave a feature request to be able to configure the VM class (not suggesting to allow multiple, just define at genesis), but it sounds like based on your original comment that this was not intended to work? I think it would be a nice feature to have for working with applications deployed with legacy VM rules e.g. ETC or just any dapp deployed before a major fork that has continuous testing.

carver:
I think the most straightforward thing to do (and the ultimate plan) would be to support configuration via genesis json. That option wasn't available before, and now all it needs is for the utility to move from trinity to py-evm

fubuloubu:
That sounds perfect. I think any use case that needs to use multiple VM versions can just run their test suite under different genesis configurations

fubuloubu:
To give an example, Vyper may end up supporting 1-3 EVM versions at some point, so our test suite would need to be run under different configurations, but we still wouldn't test upgrades because that would be too complex.

PR #191, opened 2020-05-29
Allow `py-evm` backend to use a custom VM

ralexstokes:
### What was wrong?

The `py-evm` backend was hardcoded to use the latest VM ruleset. This makes it difficult to experiment with or test future forks using this library.

### How was it fixed?

A custom VM class can now be passed to the `py-evm` backend constructor.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,7 +3,8 @@ Unreleased
 
 - Features
 
-	- Add support for gas estimate block identifiers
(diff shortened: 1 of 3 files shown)


carver:
I was going to say that this fixes #106, but I think supporting a genesis file was the conclusion there, so this would be a half-measure to get there.

Issue #147, opened 2019-01-15
Update gas limit


fubuloubu:
### What was wrong?
Current default block gas limit is 3.14m gas. Mainchain gas limit has been around 8m for a while now, which means it is possible to deploy larger contracts than what is currently testable. Lowest testnet block gas limit is 7m gas.

### How can it be fixed?

Until #129 is implemented to permanently solve the configuration issue, I would suggest that the arbitrarily-chosen default be at least 2x'd to 6.28m gas (2 Pi). This would ensure a more realistic block gas limit for testing, while retaining a sufficient margin of safety against the current limits (which can fluctuate slightly)

pipermerriam:
I'm :+1: on this.

MatthiasLohr:
Any news on that?

PR #194, opened 2020-06-15
Update default gas limit

berndbohmeier:
### What was wrong?
See #147 

### How was it fixed?
The default block gas limit was increased as suggested.

### To-Do:


kclowes:
Thanks for this @berndbohmeier! Sorry it got lost way back when. I updated it pull in master and to match the gas limit at the London hard fork. 

@fselmo do you mind taking a look?

kclowes (approved):

Issue #190, opened 2020-05-29
Improve testing of gas estimation with block identifiers


marcgarreau:
### What was wrong?
PR https://github.com/ethereum/eth-tester/pull/189 introduced support for block identifiers (e.g., "latest", "earliest") in the `estimateGas` method. A "good enough" test was included, but there's lots of room for improvement.

### How can it be fixed?
Cherry-picking from the above-mentioned PR:
@carver: 
> I guess I should have said the best kind of test would be to write a contract that burns gas in a way that's easily detectable by block number, like burning (100k * block_number) gas.

@pipermerriam: 
> Contract with a toggle-able flag which just returns immediately or burns a bunch of gas in a loop. Then you can set it and test on either side of the block number where the flag was toggled.

ackhia:
I created PR #200 to hopefully resolve this.

kclowes:
Closed with #200

PR #200, opened 2020-10-02
Block identifier gas estimates better test

ackhia:
### What was wrong?

The test for the block identifier gas estimates needed improving see #190

### How was it fixed?

I compiled the following contract:

```contract GasBurner {
    uint256 total;
    function burnBlockNumberDependentGas() public {
        for(uint256 i=0; i<block.number; i++) {
            total++; 
        }
    }
}
```

then called burnBlockNumberDependentGas with 'latest' and 'earliest' block identifiers. The results were then checked to ensure the latest estimate was bigger than the earliest.

### To-Do:

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/tools/gas_burner_contract.py
+++ b/eth_tester/tools/gas_burner_contract.py
@@ -5,13 +5,27 @@ from eth_utils import (
     function_abi_to_4byte_selector,
 )
 
+# The following contract burns gas relative to the current block number. The
+# higher the block number, the more gas is burned. It is used to test
+# functionality that uses block identifiers.
+#
+# pragma solidity >=0.4.22 <0.7.0;
+#
+# contract GasBurner {
+#     uint256 total;
+#     function burnBlockNumberDependentGas() public {
+#         for(uint256 i=0; i<block.number*10; i++) {
+#             total++;
+#         }
+#     }
+# }
 
 GAS_BURNER_BYTECODE = (
-    "6080604052348015600f57600080fd5b5060978061001e6000396000f3fe6080604052348015600f5"
+    "6080604052348015600f57600080fd5b50609b8061001e6000396000f3fe6080604052348015600f5"
     "7600080fd5b506004361060285760003560e01c8063aeecb68814602d575b600080fd5b6033603556"
-    "5b005b60008090505b43811015605f576000808154809291906001019190505550808060010191505"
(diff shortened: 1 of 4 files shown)


pipermerriam (commented):
Our latest library structure would have the gas burner contract utility living under `eth_tester.tools`

```
eth_tester/utils/gas_burner_contract.py -> eth_utils/tools/gas_burner_contract.py
```

Also it would be nice to have the source code included even if it isn't being used so that the contract can be recompiled later if needed (and for reference so that it's clear what the contract is supposed to do).

ackhia:
Hi @pipermerriam please see update. Thanks.

eth_tester/tools/gas_burner_contract.py
```diff
@@ -0,0 +1,75 @@
...
+    })
+    deploy_receipt = eth_tester.get_transaction_receipt(deploy_hash)
+    gas_burner_address = deploy_receipt['contract_address']
+    assert gas_burner_address
+    gas_burner_code = eth_tester.get_code(gas_burner_address)
+    assert len(gas_burner_code) > 2
+    return gas_burner_address
+
+
+def _make_call_gas_burner_transaction(eth_tester, contract_address, fn_name, fn_args=None):
```
pipermerriam, line 56: The `None` default argument for `fn_args` could actually just be the empty tuple here in the function signature since tuples are immutable.  That would allow removal of the `if fn_args is None: ...` block a few lines down.

eth_tester/tools/gas_burner_contract.py
```diff
@@ -0,0 +1,75 @@
...
+    from eth_abi import encode_abi
```
pipermerriam, line 57: What is the reason for this import being here inside the function body instead of at the top of the module?

pipermerriam (approved):

eth_tester/tools/gas_burner_contract.py
ackhia: Good spot. I copied this code from math_contract.py so I'm not sure why it was done this way. I'll put in a default tuple anyway.

eth_tester/tools/gas_burner_contract.py
ackhia: Again this was from copied code. Maybe it was for performance reasons. I'll move it up top anyway.

ackhia:
All done @pipermerriam

kclowes:
Thanks @ackhia! Sorry this got lost. I pulled in the most recent master, and will merge once tests pass!

PR #201, opened 2020-10-08
Relax upper bound of RLP dependency

cburgdorf:
### What was wrong?

The current RLP dependency has a hard bound on `2.0.0a1`. Trinity, Py-EVM and friends need to migrate to `2.0.0a2` so we need to update the dependency. 

### How was it fixed?

Changed it to `"rlp>=1,<3"` which is  inline with what [`eth-enr` does](https://github.com/ethereum/eth-enr/blob/master/setup.py#L64) and causes less cascading releases whenever we cut a new alpha release of RLP.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         "eth-abi>=2.0.0b4,<3.0.0",
         "eth-keys>=0.2.1,<0.4.0",
         "eth-utils>=1.4.1,<2.0.0",
-        "rlp>=1.1.0,<=2.0.0-alpha.1",
+        "rlp>=1.1.0,<3",
         "semantic_version>=2.6.0,<3.0.0",
     ],
     extras_require=extras_require,


carver:
Same as https://github.com/ethereum/py-trie/pull/119

pipermerriam (approved):
:+1: 

have you confirmed whether the beta gets installed with this syntax or does it move the library back to the `1.x` line by default unless something else explicitely specifies the alpha/beta versions?

cburgdorf:
Mmh..ok, trying to install `pip install -e .[dev]` on this PR leaves me with:

>ERROR: py-evm 0.3.0a19 has requirement rlp==2.0.0-alpha.1, but you'll have rlp 1.2.0 which is incompatible.

I would have assumed it should get along with py-evm requiring `rlp==2.0.0-alpha.1`. I mean, I'm fine with all these libraries defaulting to the `1.x` line and only the applications (such as Trinity) pinning to a beta version. But this error suggests there being trouble ahead because Trinity would basically do the same thing than what it complains about..hmpf..

cburgdorf:
I played with the different syntax but the only two options I see are:

1. We keep doing what we did and pin it everywhere to `rlp==2.0.0-beta.1` at the cost of having to make a bunch of releases in tandem if we have to cut new rlp releases

2. We cut rlp 2.0.0 stable.

Personally, I don't see an issue 2.). If we really want to make other breaking changes we would cut a version 3. which was our intend anyway by already cutting a 2b1.

pipermerriam:
Pinning to exact versions is going to be problematic because of the ubiquity of the library.  I'm ok cutting a stable release.  It seems like the lesser evil here.

cburgdorf:
>Pinning to exact versions is going to be problematic because of the ubiquity of the library. I'm ok cutting a stable release. It seems like the lesser evil here.

Yes, I agree. Pinging @carver for another pair of eyes but if no one objects, I'll take the following actions tomorrow:

1. Cut rlp 2 stable
2. Merge the PRs with the upper bound set to `<3`
3. Cut new releases accordingly

carver:
> 2\. We cut rlp 2.0.0 stable.
> 
> Personally, I don't see an issue 2.). If we really want to make other breaking changes we would cut a version 3. which was our intend anyway by already cutting a 2b1.

So how confident are we that `rlp` v2 works cleanly on other architectures? IIRC, now it does not try to install the rust library by default, so it should be working just as cleanly as v1, yeah? If that's right (and has been tested on an architecture that doesn't support the rust lib) then :+1: for a v2 stable release.

---


> I would have assumed it should get along with py-evm requiring `rlp==2.0.0-alpha.1`.

Did you try it with the `--use-feature=2020-resolver` flag? That will be the default in pip _soon_ :tm:.

cburgdorf:
> IIRC, now it does not try to install the rust library by default, so it should be working just as cleanly as v1, yeah?

Correct, It just defines an `extra[rust-backend]` and has the code to use `rusty-rlp` if it is importable. Other than than that, it's pretty much v1.

>Did you try it with the pip install --use-feature=2020-resolver flag? That will be the default in pip soon tm.

Good point! The flag does the trick indeed. With that flag, this PR would install cleanly without complaining and resolve to `rlp==2.0.0a1` (as pinned by Py-EVM).

But I wonder, does that change anything in the short term?

carver:
> But I wonder, does that change anything in the short term?

Yeah, I've been considering it okay if a library/project only installs using the new resolver.

carver (approved):

cburgdorf:
>Yeah, I've been considering it okay if a library/project only installs using the new resolver.

Good to know. I think I would be game to go that route if it weren't for such an ubiquitous library. We'd probably get lots of issues from user with installation issues which we then have to guide through upgrading pip and using that flag. Most of the installation complaints came through web3.py users so far but there are a bunch of libraries that are affected.

I'm going to stick to the plan of cutting a stable v2 but its great to see that pips dependency resolving is getting better!

carver:
FWIW, the default is supposed to come very soon:

> We are preparing to change the default dependency resolution behavior and make the new resolver the default in pip 20.3 (in October 2020).
[source](https://discuss.python.org/t/announcement-pip-20-2-release/4863)

cburgdorf:
Yep, I was reading that. But if I understand that correctly there's nothing that will enforce people to switch to pip 20.3 this month. People could happily stay with `20.2`, no? Or will that one stop working...I didn't read anything like that.

carver:
Right, it's just easy to say that you have to update to the latest pip to get any of our libraries working correctly. Even easier than saying to use the `--use-feature=2020-resolver`

PR #204, opened 2021-03-24
Update py-evm for Berlin support

carver:
### What was wrong?

Using a pre-Berlin evm.

### How was it fixed?

Update py-evm. Set default VM to Berlin. Resolve a couple changed APIs.

### To-Do:

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -156,8 +156,8 @@ def setup_tester_chain(
     from eth.db import get_db_backend
 
     if vm_configuration is None:
-        from eth.vm.forks.muir_glacier import MuirGlacierVM
-        no_proof_vms = ((0, MuirGlacierVM.configure(consensus_class=NoProofConsensus)),)
+        from eth.vm.forks import BerlinVM
+        no_proof_vms = ((0, BerlinVM.configure(consensus_class=NoProofConsensus)),)
     else:
         consensus_applier = ConsensusApplier(NoProofConsensus)
         no_proof_vms = consensus_applier.amend_vm_configuration(vm_configuration)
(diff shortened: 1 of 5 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -461,8 +459,7 @@ def _get_normalized_and_signed_evm_transaction(self, transaction, block_number='
 
     def send_raw_transaction(self, raw_transaction):
         vm = _get_vm_for_block_number(self.chain, "latest")
-        TransactionClass = vm.get_transaction_class()
-        evm_transaction = rlp.decode(raw_transaction, TransactionClass)
+        evm_transaction = vm.get_transaction_builder().decode(raw_transaction)
```
carver, line 462: This API changed, because not all raw transactions are valid RLP anymore...

eth_tester/utils/backend_testing.py
```diff
@@ -284,9 +284,16 @@ def test_send_raw_transaction_valid_raw_transaction(self, eth_tester, is_pending
         # assert that the raw transaction is confirmed and successful
         assert receipt['transaction_hash'] == transaction_hash
 
-    def test_send_raw_transaction_invalid_raw_transaction(self, eth_tester):
+    def test_send_raw_transaction_invalid_rlp_transaction(self, eth_tester):
         self.skip_if_no_evm_execution()
         invalid_transaction_hex = '0x1234'
+        import eth
+        with pytest.raises(eth.exceptions.UnrecognizedTransactionType):
+            eth_tester.send_raw_transaction(invalid_transaction_hex)
```
carver, line 292: Maybe this should wrap the exception in something like a `ValidationError`, although I think we've leaned pretty far away from actually supporting multiple backends.

marcgarreau: yeah, don't think we have the bandwidth to make multiple backends a priority either.

marcgarreau (approved):
🚢

PR #205, opened 2021-04-09
add python 3.9 support

marcgarreau:
### What was wrong?

A py-evm update is required to bring eth-tester and web3.py (https://github.com/ethereum/web3.py/issues/1769) into the 3.9 age.

### To-Do:


kclowes (commented):
LGTM! 🚀

PR #206, opened 2021-09-20
London support

fselmo:
### What was wrong?

- Support for London is needed. ~~This cannot be done until py-evm changes are in~~ [they're in!].
- Support for access list + access list transactions was needed (Berlin)

### How was it fixed?

- `py-evm` version upgraded to `v0.5.0-alpha.1`
- London support was added
   - Dynamic fee transaction support: `type`='0x2', `max_fee_per_gas`, `max_priority_fee_per_gas`, `access_list`, `y_parity`
   - `base_fee_per_gas` support added to blocks
- Access list support was added (Berlin)
   - Access list transaction support: `type`='0x1', `access_list`, `chain_id`, `y_parity`
- Support for transaction params `chain_id`, `type`, `y_parity` for typed transactions
- Support for receipt params `type`, `effective_gas_price`
- Tests added

### To-Do:
- [X] Add support for new transaction params (access list, max fees, transaction type, chain id?)
- [X] Add support for new block field `base_fee_per_gas`
- [X] Get tests to pass locally
- [X] When tests pass, make sure everything is actually implemented appropriately + clean up rough initial commit
- [X] Validate new fields appropriately
- [X] Add new tests
- [x] Clean up commit history after PR is approved

#### Cute Animal Picture

IMG_20211019_093537


eth_tester/utils/backend_testing.py
```diff
@@ -862,6 +846,16 @@ def test_receipt_effective_gas_price_for_mined_transaction_max_fee_limit(self, e
         # effective gas price
         assert receipt['effective_gas_price'] == max_fee
 
+    def test_get_transaction_receipt_for_unmined_transaction_raises(self, eth_tester):
```
fselmo, line 849: This move successfully corrected git's algorithm and didn't make it look like I re-wrote over this test (saved the old commit message)

eth_tester/validation/common.py
```diff
@@ -179,6 +179,16 @@ def validate_array(value, validator):
         raise ValidationError(error_message)
 
 
+def validate_transaction_type(value):
```
fselmo, line 182: I chose to validate here based on whether or not the value validates to `0`, `1`, or `2` integers. This is because `0x01` is the same as `0x1` and both should be accepted. Internally, we only use `0x0`, `0x1` and `0x2` for the type and this is because we pop the type out of the transaction (after we validate it here) and instead add the type that is appropriate based on the transaction parameters alone.

We do this because we don't want `type` to be a required field to send a transaction.

marcgarreau (approved):
Big thanks for the whopper 🍔 of a changeset. Looks good from IRL review, and just did some more smoke tests locally with the relevant branches of py-evm and web3.py. Just wanted to document one UX bit from that testing:

If building a transaction manually and you only include one of the 1559 keys, `w3.eth.account.sign_transaction(tx, acct.address)` will raise from within eth-account, having decided it's a legacy transaction without a `gasPrice`. 

(`w3.eth.send_transaction` provides cleaner error messages, e.g., `web3.exceptions.InvalidTransaction: maxPriorityFeePerGas must be defined in a 1559 transaction.`)

kclowes (approved):
LGTM!! 🎉 🚢 :shipit:

eth_tester/backends/mock/serializers.py
```diff
@@ -42,11 +42,14 @@ def serialize_full_transaction(transaction, block, transaction_index, is_pending
...
         return serialized_transaction
     else:
-        return assoc(
-            serialized_transaction,
-            'gas_price',
-            calculate_effective_gas_price(transaction, block)
+        # TODO: Sometime in 2022 the inclusion of gas_price may be removed from dynamic fee
+        #  transactions and we can get rid of this behavior.
+        #  https://github.com/ethereum/execution-specs/pull/251
+        gas_price = (
```
fselmo, line 48: This was a whoopsie before that I caught when testing via web3.py. If the transaction is pending, `gas_price` should be equal to `max_fee_per_gas` for dynamic fee transactions, otherwise (if already mined) it is the effective gas price. The explanation is in the link in the TODO.

PR #209, opened 2021-10-29
feat: support mnemonic when generating genesis accounts

unparalleled-js:
### What was wrong?

I need to be able to control which accounts get funded. The way I have done this with other providers is via mnemonic. I was hoping we can add mnemonic support here as well?  Or at least, a way to set the private keys? I am able to set genesis data for funding accounts, but I can't actually make eth-tester use those accounts.

### How was it fixed?

Added `mnemonic` parameter and handle it by using tooling from `eth-account` to generate accounts that way instead of using the default.

### To-Do:

#### Cute Animal Picture

baby_turtle

### DIFF
--- a/README.md
+++ b/README.md
@@ -817,9 +817,25 @@ Then pass the generated `custom_genesis_params` `dict` to the backend's `__init_
 >>> t = EthereumTester(backend=pyevm_backend)
 ```
 
-Overriding genesis state is similar to overriding genesis parameters but requires the consideration of test accounts.
-To override the genesis state of accounts, pass a `state_overrides` `dict` to `PyEVMBackend._generate_genesis_state`,
-and optionally, the number of accounts to create.  
+Similarly to `genesis_parameters`, override the genesis state by passing in an `overrides` `dict`
+to `PyEVMBackend._generate_genesis_state`. Optionally, provide `num_accounts` to set the number of accounts.
+
+For more control on which accounts the backend generates, use the `from_mnemonic()` classmethod. Give it
+a `mnemonic` (and optionally the number of accounts) and it will use that information to generate the accounts.
+Optionally, provide a `genesis_state_overrides` to adjust the `genesis_state`.
+```python
+>>> from eth_tester import PyEVMBackend, EthereumTester
+>>> from eth_utils import to_wei
+>>> from hexbytes import HexBytes
+>>>
+>>> pyevm_backend = PyEVMBackend.from_mnemonic(
+>>>    'test test test test test test test test test test test junk',
+>>>    genesis_state_overrides={'balance': to_wei(1000000, 'ether')}
+>>> )
+>>> t = EthereumTester(backend=pyevm_backend)
+>>> print(t.get_accounts()[0])  # Outputs 0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C
+>>> print(t.get_balance('0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C'))  # Outputs 1000000000000000000000000
+```
 
 *NOTE: The same state is applied to all generated test accounts.* 
 
--- a/README.md
+++ b/README.md
@@ -822,7 +822,7 @@ to `PyEVMBackend._generate_genesis_state`. Optionally, provide `num_accounts` to
 
 For more control on which accounts the backend generates, use the `from_mnemonic()` classmethod. Give it
 a `mnemonic` (and optionally the number of accounts) and it will use that information to generate the accounts.
-Optionally, provide a `genesis_state_overrides` to adjust the `genesis_state`.
+Optionally, provide a `genesis_state_overrides` `dict` to adjust the `genesis_state`.
 ```python
 >>> from eth_tester import PyEVMBackend, EthereumTester
 >>> from eth_utils import to_wei
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -9,6 +9,8 @@ from eth_abi.exceptions import (
     DecodingError
 )
 
+from eth_account.hdaccount import HDPath, seed_from_mnemonic
+
 from eth_utils import (
     encode_hex,
     int_to_big_endian,
@@ -111,6 +113,17 @@ def get_default_account_keys(quantity=None):
         yield private_key
 
 
+@to_tuple
+def get_account_keys_from_mnemonic(mnemonic, quantity=None):
+    keys = KeyAPI()
+    seed = seed_from_mnemonic(mnemonic, "")
+    quantity = quantity or 10
+    for i in range(0, quantity):
+        hd_path = HDPath(f"m/44'/60'/0'/{i}")
+        private_key = keys.PrivateKey(hd_path.derive(seed))
+        yield private_key
+
+
 @to_dict
 def generate_genesis_state_for_keys(account_keys, overrides=None):
     for private_key in account_keys:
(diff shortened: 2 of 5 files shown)


eth_tester/backends/pyevm/main.py
```diff
@@ -259,7 +276,13 @@ def _get_vm_for_block_number(chain, block_number):
 class PyEVMBackend(BaseChainBackend):
     chain = None
 
-    def __init__(self, genesis_parameters=None, genesis_state=None, vm_configuration=None):
+    def __init__(
+        self,
+        genesis_parameters=None,
+        genesis_state=None,
```
unparalleled-js, line 282: I noticed none of these params are documented besides `vm_configuration`...

If it would be helpful, I can document this method's params and maybe a bit to the README about the `mnemonic` usage?

kclowes: That would be very helpful, thanks!

eth_tester/backends/pyevm/main.py
```diff
@@ -146,7 +159,8 @@ def setup_tester_chain(
         genesis_params=None,
         genesis_state=None,
         num_accounts=None,
-        vm_configuration=None):
+        vm_configuration=None,
+        mnemonic=None):
```
unparalleled-js, line 163: Even though it is a key value arg, I always feel safer appending new arguments to the end of the signature to lessen the chance of breaking people's stuff...

however, I noticed that these are alphabetized.. Is that intentional?

README.md
```diff
@@ -819,7 +819,30 @@ Then pass the generated `custom_genesis_params` `dict` to the backend's `__init_
 
 Overriding genesis state is similar to overriding genesis state, but requires the consideration of test accounts.
 To override the genesis state of accounts, pass a `state_overrides` `dict` to `PyEVM.generate_genesis_state`,
-and optionally, the number of accounts to create.
```
unparalleled-js, line 822: These changes will look better once https://github.com/ethereum/eth-tester/pull/207 merges

kclowes (commented):
I ran out of time to do a full in-depth review today. Will pick up again on Monday! Thanks!

README.md
```diff
@@ -817,9 +818,28 @@ Then pass the generated `custom_genesis_params` `dict` to the backend's `__init_
 >>> t = EthereumTester(backend=pyevm_backend)
 ```
 
-Overriding genesis state is similar to overriding genesis parameters but requires the consideration of test accounts.
-To override the genesis state of accounts, pass a `state_overrides` `dict` to `PyEVMBackend._generate_genesis_state`,
-and optionally, the number of accounts to create.  
+To override the genesis state of accounts, pass a `state_overrides` `dict` to `PyEVMBackend._generate_genesis_state`.
```
unparalleled-js, line 821: I would really prefer to call a public method.

Can we make the same method except public that passthroughs to the protected one? That way, anyone who followed this readme won't have broken stuff and new users, like myself, can use the public one.

It does not have to happen on this PR but it would be great and I can take care of it

unparalleled-js: https://github.com/ethereum/eth-tester/pull/210

^ If this merges first, I will adjust this PR accordingly.

kclowes: Yep, that sounds good to me!

README.md
```diff
@@ -817,9 +817,29 @@ Then pass the generated `custom_genesis_params` `dict` to the backend's `__init_
...
+>>> # The following address stems from the mnemonic
+>>> address = '0x1e59ce931B4CFea3fe4B875411e280e173cB7A9C'
+>>> mnemonic = 'test test test test test test test test test test test junk'
+>>>
+>>> genesis_state = PyEVMBackend._generate_genesis_state(
+>>>   overrides={'balance': to_wei(1000000, 'ether')},
+>>>   mnemonic=mnemonic,
+>>>   num_accounts=1
+>>> )
+>>> pyevm_backend = PyEVMBackend(genesis_state=genesis_state, mnemonic=mnemonic)
```
kclowes, line 838: I may be misunderstanding something, but I ran through this example locally and it feels a little bit clunky that you have to pass in `mnemonic` here and in the `genesis_state`. Is there a way to simplify and only pass it with `genesis_state`? Or is that a check to make sure the `mnemonic` is the same?

eth_tester/backends/pyevm/main.py
```diff
@@ -111,6 +113,17 @@ def get_default_account_keys(quantity=None):
         yield private_key
 
 
+@to_tuple
+def get_account_keys_from_mnemonic(mnemonic, quantity=None):
```
kclowes, line 117: nit - could this be:
```suggestion
def get_account_keys_from_mnemonic(mnemonic, quantity=10):
```
And then we could get rid of line 119?

kclowes (commented):
This is looking really good, thank you! I left a few comments that I'd be interested to get your opinion on. Then, I think the only thing missing is a few tests! It looks like [here-ish](https://github.com/ethereum/eth-tester/blob/master/tests/backends/test_pyevm.py#L99) might be a good place to add them. I'd like to see a test similar to the example you lay out in the README. Maybe make sure the accounts are the accounts you expect and the number of accounts is correct if you pass in num_accounts vs not. I don't see anything else that really sticks out to me as far as testing, but feel free to add more if you think of anything else that could go wrong. Thanks again!

eth_tester/backends/pyevm/main.py
unparalleled-js: Yes, I was copying the method defined above it; I can change them both since they are effectively the same thing

eth_tester/backends/pyevm/main.py
```diff
@@ -9,6 +9,8 @@
     DecodingError
 )
 
+from eth_account.hdaccount import HDPath, seed_from_mnemonic
```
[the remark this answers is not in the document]
unparalleled-js: In `setup.py`, I pinned the version of `eth-account`. This way, if the next version of `eth-account` has breaking changes, it will be opt-in only; we can upgrade as we go.  

However, another idea I had was to allow a private key generator. This way, users pass in a generator that generates private keys up to a certain the given `num_accounts` or something.

I would prefer the `mnemonic` method because it makes it similar to other testing-providers and but am good with anything!

README.md
unparalleled-js: The best solution I found is to create a helper factory classmethod `from_mnemonic` that does both... It is tricky because both use the private keys for different purposes and i couldn't really find a way to get rid of one and not the other. I really like the class method approach though! (You will see when I push). Now, it is even simpler to do this and the user only needs to pass the mnemonic once.

eth_tester/backends/pyevm/main.py
unparalleled-js: Ah when running the tests, I see why they did it this way... It is because sometimes this is kwarg passthrough in which case `None` gets passed in. So the line `quantity = quantity or 10` handles the case when someone does `quantity=None`

kclowes: Ah, interesting. I think I slightly prefer leaving it as None then, but I'll leave it up to you!

unparalleled-js: I change it back to `None` !

setup.py
```diff
@@ -54,6 +54,7 @@
     include_package_data=True,
     install_requires=[
         "eth-abi>=2.0.0b4,<3.0.0",
+        "eth-account==0.5.6",
```
kclowes, line 57: Let's open this up to be non-breaking changes. Since it's one the snake-charmers maintain, I'm relatively confident it won't break unless be bump the minor version.
```suggestion
        "eth-account>=0.5.6,<0.6.0",
```
It will help to play nice with our other libraries.

kclowes (commented):
One last comment about opening up the eth-account version, and then this looks good to me. Thank you!

kclowes (approved):
Thank you!

kclowes:
Will do a release once the London changes are in. Hopefully this week!

PR #210, opened 2021-10-31
chore: create public versions of methods

unparalleled-js:
### What was wrong?

The README recommends using protected methods. It is difficult to know if this method is considered public or an implementation detail. I would like to be able to call these methods and not worry that they may break at some point.

### How was it fixed?

I made "public" classmethods that passthrough to the original static methods. This way, users who may be calling the protected methods won't have their stuff break.

### To-Do:

#### Cute Animal Picture
baby_bear

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,10 +1,7 @@
-v0.6.0-beta.1
+v0.6.0-beta.1 (Unreleased)
 -------------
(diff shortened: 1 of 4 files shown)


kclowes (approved):
Looks good to me! Thank you!

Issue #212, opened 2021-11-04
Import error when installing update 0.6.0b1


unparalleled-js:
* Version: 0.6.0b1
* Python: 3.9.7
* OS: macOS

### What was wrong?

Crashes when trying to import the backend.

```python
from eth_tester.backends import PyEVMBackend
```

Gets the issue:

```
ModuleNotFoundError: No module named 'eth.vm.forks.berlin'
```

My pyevm version is 0.3.0a20, which I see is not supported! However, I already had it installed in my environment from a separate package and there is nothing in this setup.py that gets mad about that.

I don't believe `eth.vm.forks.berlin` existed until 0.4.0, correct me if I am wrong!
Since py-evm is not listed in the `install_requires`, it does not ensure it gets updated or anything.

### How can it be fixed?

Before importing `BaseTransaction` and `TypedTransaction`, check if pyevm version is high enough

unparalleled-js:
Ah, I guess this is solved on my end by including `"eth-tester[pyevm]",` in the setup.py.

At the same, checking the version in `is_pyevm_available()` is a little more clear I think

PR #213, opened 2021-11-04
fix: unable to import PyEVMBackend from fresh upgrade

unparalleled-js:
### What was wrong?

fixes: #212 

### How was it fixed?

Check version when checking if py-evm installed

### To-Do:

#### Cute Animal Picture

baby_squirrel


kclowes (commented):
Thanks for the change, you're so fast! This is sort of a nit, but the error messaging where `is_pyevm_available` gets called assumes that pyevm isn't installed (like [here](https://github.com/ethereum/eth-tester/blob/42462659fbb5b2f13c9701ef30ee2712af6632e1/eth_tester/backends/pyevm/main.py#L302), and [here](https://github.com/ethereum/eth-tester/blob/42462659fbb5b2f13c9701ef30ee2712af6632e1/eth_tester/backends/__init__.py#L30)), so I wonder if it would be better to make these two separate checks? Like leave `is_pyevm_available`, and then add a method that checks that the version of py-evm is a high enough version so we can make the error messages slightly better. Or we can update the wording of the error messages. What do you think?

unparalleled-js:
> here

I think what I would do then is change to the error messages to indicate `pyevm >= 5.0.0` is not installed and maybe change the method name to `is_supported_pyevm_version_available()`. I think that would be easier because we will always have to check both anyway.

kclowes:
@unparalleled-js - This is ready to go, right?

kclowes (approved):

unparalleled-js:
Yeah, unless you know a way to dynamically get the min supported version. I tried to see if we could do that but couldn't find a way.

kclowes:
For [this line](https://github.com/ethereum/eth-tester/pull/213/files#diff-174d7e8e64979634cd2867dce9771d4afe3646617ffde84cd4c90336e4dea669R22)/error messages? Is it set somewhere else?

kclowes:
I felt bad for breaking your CHANGELOG again and again, so I added towncrier to eth-tester. So now there will be newsfragments like our other repos, and less merge conflicts :tada:. You can pull in master to get that update. I'm happy to add the snippet from the CHANGELOG in this PR to a newsfragment if you want, just let me know!

eth_tester/backends/pyevm/utils.py
```diff
@@ -17,10 +17,6 @@ def get_pyevm_version():
...
-def is_pyevm_available():
-    pyevm_version = get_pyevm_version()
-
-    if pyevm_version is None:
-        return False
-    else:
-        return True
+def is_supported_pyevm_version_available():
+    version = get_pyevm_version()
+    return version and version >= Version("0.5.0")
```
unparalleled-js, line 22: This is the version that would be nice to dynamically get.
I can think of a way to do it but that involves creating a source-of-truth that both here and `setup.py` can use... However, it seems like there should be a way using `pkg_resources` or something but I could not figure out a way.

unparalleled-js:
> I felt bad for breaking your CHANGELOG again and again, so I added towncrier to eth-tester. So now there will be newsfragments like our other repos, and less merge conflicts 🎉. You can pull in master to get that update. I'm happy to add the snippet from the CHANGELOG in this PR to a newsfragment if you want, just let me know!

I think I upgraded to the newsfragments approach correctly!

eth_tester/backends/pyevm/utils.py
kclowes: Yeah, I don't see a nice way to do it either. I think this is good enough for now 👍

PR #214, opened 2021-11-04
refactor: use public methods

unparalleled-js:
### What was wrong?

Updating references to the new public methods now that they are available.
The one that is most important is the one in the README.

### How was it fixed?

Removing underscore prefix!

### To-Do:

#### Cute Animal Picture

baby_bear


kclowes (approved):
Not sure how I missed this one, but looks good to me! Feel free to @ me if I miss something! Thanks @unparalleled-js!

PR #215, opened 2021-11-05
Allow nonce for estimate eth_estimateGas and eth_call

fselmo:
### What was wrong?

- If a transaction has a `nonce` parameter and a call is made to `eth_estimateGas` or `eth_call`, this raises an exception in eth-tester when validating the inbound transaction. Though the `nonce` isn't used at all, it shouldn't necessarily raise here and indeed it creates issues within core tests in web3.py using the pyevm backend.

This isn't an issue with clients like geth so it shouldn't be an issue for eth-tester either. 

### How was it fixed?

- Allow `nonce` for `eth_estimateGas` and `eth_call` as this is indeed a valid transaction parameter

### To-Do:

- [X] Add entry to the [CHANGELOG](https://github.com/ethereum/eth-tester/blob/master/CHANGELOG)

#### Cute Animal Picture

IMG_20211105_121059


kclowes (approved):
LGTM! 🚀

Issue #216, opened 2021-11-08
Unable to use BerlinVM configuration


unparalleled-js:
* Version: 0.6.0b1
* Python: 3.9.7
* OS: macOS

### What was wrong?

When I create a backend using `BerlinVM`, like this:

```python
from eth.consensus import NoProofConsensus
from eth.vm.forks import BerlinVM
from eth_tester.backends import PyEVMBackend

mnemonic = "test test test test test test test test test test test junk"
vm_configuration = ((0, BerlinVM.configure(consensus_class=NoProofConsensus)),)
tester = PyEVMBackend.from_mnemonic(
    mnemonic, vm_configuration=vm_configuration
)
```

and then I try to send a transaction (`web3.eth.send_raw_transaction()`), I get this traceback:

```
 File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/module.py", line 58, in caller
    result = w3.manager.request_blocking(method_str, params, error_formatters)
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/manager.py", line 154, in request_blocking
    response = self._make_request(method, params)
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/manager.py", line 133, in _make_request
    return request_func(method, params)
  File "cytoolz/functoolz.pyx", line 250, in cytoolz.functoolz.curry.__call__
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/middleware/formatting.py", line 76, in apply_formatters
    response = make_request(method, params)
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/middleware/gas_price_strategy.py", line 34, in middleware
    return make_request(method, params)
  File "cytoolz/functoolz.pyx", line 250, in cytoolz.functoolz.curry.__call__
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/middleware/formatting.py", line 74, in apply_formatters
    response = make_request(method, formatted_params)
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/middleware/attrdict.py", line 33, in middleware
    response = make_request(method, params)
  File "cytoolz/functoolz.pyx", line 250, in cytoolz.functoolz.curry.__call__
  File "/Users/jules/virtualenvs/harambe/lib/python3.9/site-packages/web3/middleware/formatting.py", line 76, in apply_formatters
[...]

PR #218, opened 2021-11-10
Don't add base_fee to block serializer until London

kclowes:
### What was wrong?

Closes issue #216 - Base fee per gas was being added for all blocks, even those prior to London, which was causing an AttributeError.

### How was it fixed?
If the block is London or above, add `base_fee_per_gas` to the block attributes. Otherwise, leave it off. 

### To-Do:

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,11 @@
+Unreleased
+----------
+
+- Bugfixes
+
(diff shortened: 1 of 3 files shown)


fselmo (approved):
LGTM! Thanks for the update!

PR #224, opened 2021-11-12
Return `v` instead of `y_parity` for typed transactions:

fselmo:
### What was wrong?

- JSON-RPC responses across clients don't generally use `y_parity` nomenclature for typed transactions. They instead keep the `v` signature parameter with the `y_parity` value. Recent London changes return the `y_parity` as is. 

### How was it fixed?

- Don't return `y_parity` but instead return `v` with the `y_parity` value from the signed transaction.

### To-Do:

- [X] Add entry to the [CHANGELOG](https://github.com/ethereum/eth-tester/blob/master/CHANGELOG)
- [x] Validate `v` appropriately for typed transactions (may only be 0 or 1)

#### Cute Animal Picture

IMG_20211111_170635


kclowes (approved):
LGTM! Some nice cleanups in there! 🚢

PR #227, opened 2021-11-19
Support for pre-London blocks missing a base fee during validation

fselmo:
### What was wrong?

- We had removed ``base_fee_per_gas`` serialization from pre-London blocks but we were still validating all blocks equally. This would break pre-London blocks on a simple `get_block()` call.

### How was it fixed?

- Set ``base_fee_per_gas`` value to ``None`` during validation if the param is missing (pre-London blocks). Pop this value back out during block normalization. This returns a block without a ``base_fee_per_gas`` for pre-London blocks which seems to be the default behavior in other clients as well.

### To-Do:

### DIFF
--- a/eth_tester/normalization/outbound.py
+++ b/eth_tester/normalization/outbound.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+from toolz import dissoc
+
 from eth_utils.curried import (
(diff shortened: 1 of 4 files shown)


kclowes (approved):
LGTM! Nice catch!

PR #228, opened 2021-12-26
Add missing fields to transaction receipt

fselmo:
### What was wrong?

* Receipts are missing the `to` and `from` keys as well as some others. We will also be able to test this via web3 [here](https://github.com/ethereum/web3.py/blob/a6642a22496faa76d0f1a900b80d87e811b31987/tests/integration/test_ethereum_tester.py#L418).

### How was it fixed?

* Add `to`, `from`, and `status` to receipts.
* Add tests

### To-Do:

#### Cute Animal Picture

Cute animal picture


eth_tester/validation/outbound.py
```diff
@@ -173,17 +173,26 @@ def _validate_outbound_access_list(access_list):
 )
 
 
+def validate_status(value):
```
kclowes, line 176: I like this change 😎

kclowes (approved):
LGTM! :shipit:

PR #229, opened 2022-01-04
fix: update pin of py-evm to stop typing-extensions pinning issues

fubuloubu:
### To-Do:

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ extras_require = {
     'py-evm': [
         # Pin py-evm to exact version, until it leaves alpha.
         # EVM is very high velocity and might change API at each alpha.
-        "py-evm==0.5.0a1",
+        "py-evm==0.5.0a2",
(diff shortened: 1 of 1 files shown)


fubuloubu:
Ugh and `eth-account` needs to update pin of `eth-keys`

kclowes:
Yeah, this is high on my list @fubuloubu. We have to do it down the whole chain.

fubuloubu:
> Yeah, this is high on my list @fubuloubu. We have to do it down the whole chain.

Thanks! Sorry about opening these PRs, was hoping that it'd be easier, but at least we can keep track of the conversation a bit more. We have a workaround by forcing `type-extensions` to get docker working but ofc would rather avoid that.

kclowes:
No worries, thanks for taking initiative! I was out for the holidays and am just now catching up, so you should see some progress here shortly. I'm glad to hear you have a workaround in the meantime!

kclowes:
Closing in favor of #230. Thanks again!

PR #231, opened 2022-01-20
Drop Python 3.6 support, add Python 3.10 support

kclowes:
### What was wrong?
Adds Python 3.10 support, drops python 3.6 support. This is needed for our other dependency upgrades.

### To-Do:

#### Cute Animal Picture

Cute animal picture


fselmo (approved):
📈 lgtm!

PR #235, opened 2022-05-20
chore: upgrade eth dependencies

unparalleled-js:
### What was wrong?

I need to upgrade my eth dependencies

### How was it fixed?

I have to first upgrade this https://github.com/ethereum/py-evm/pull/2062
and then maybe include that it in the py-evm dependencies as part of this PR

### To-Do:

#### Cute Animal Picture

Cute animal picture


kclowes:
Thanks for the PR! See my comment [here](https://github.com/ethereum/py-evm/pull/2062#issuecomment-1133075193)

kclowes:
This should be fixed now in a more sensible way. Lmk if you're still having dependency problems though, and we'll prioritize fixing!

PR #238, opened 2022-08-22
Breaking Dependency Changes

kclowes:
### What was wrong?

Need to upgrade the dependencies and issue a sem-versioned breaking release

### How was it fixed?

Cherry-picked the correct commit, upgraded py-evm since that has the right sem version now, and upgraded towncrier. I could totes pull out the towncrier version into another PR if anyone feels strongly. I also made sure all these dependencies install in web3 v6.

### To-Do:

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,13 +32,13 @@ test-all:
 
 validate-docs:
 	python newsfragments/validate_files.py
-	towncrier --draft
+	towncrier build --draft
 
-notes:
+notes: validate-docs
 	# Let UPCOMING_VERSION be the version that is used for the current bump
 	$(eval UPCOMING_VERSION=$(shell bumpversion $(bump) --dry-run --list | grep new_version= | sed 's/new_version=//g'))
 	# Now generate the release notes to have them included in the release commit
-	towncrier --yes --version $(UPCOMING_VERSION)
+	towncrier build --yes --version $(UPCOMING_VERSION)
 	git commit -m "Compile release notes"
 
 release: clean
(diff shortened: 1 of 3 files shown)


fjarri:
Is it still a draft? I would very much like to see this released.

Also, copying a note from a similar PR I opened without noticing this one:

Having to do this is a consequence of a bigger problem, namely using upper bounds in dependencies. I suspect this was initially caused by bringing development practices from JS into Python. In JS this works because the every dependency has its own subdependencies which do not intersect; in Python the dependency namespace is flat, and upper bounds cause more problems than they solve (the fact no code had to be changed after bumping the bounds illustrates that). In this PR I tried to follow the general scheme of eth-* packages to set the new upper bounds.

kclowes:
@fjarri Thanks for bringing this up. I've been looking into capping dependencies or not, this morning and I'll chat with the team to figure out if this is the path we want to take. The biggest con I see is that if there is a dependency upgrade and it breaks (like `websockets` v11 in web3.py), we have to drop everything and either temporarily cap or fix it so that CI can pass, which would be blocking to the whole team while someone is getting that fix out. It's not a huge deal, but does take time to context switch, etc.

pacrob (approved):
all lgtm, no need for separate PR

fjarri:
@kclowes it is a trade-off for sure, there's no perfect solution, unfortunately. But in my experience not having upper bounds results in much fewer problems overall than having them, as long as the package is maintained. Also linking a very informative article about the issue, which may be of interest: https://iscinumpy.dev/post/bound-version-constraints/

PR #240, opened 2022-08-25
Bump py-evm version and related dependencies

fjarri:
### What was wrong?

`py-evm` was pinned to an old version, which, in turn, held back all the related `eth-*`dependencies. Py-evm 0.6.0a1 has been released recently which bumped all of them, so finally they can be bumped here as well.

### How was it fixed?

- `py-evm` version bumped to `0.6.0a1`
- `eth-abi` bumped to `>=3.0.0,<4.0.0`
- `eth-account` bumped to `>=0.6.0,<0.8.0`
- `eth-keys` bumped to `>=0.4.0,<0.5.0`
- `eth-utils` bumped to `>=2.0.0,<3.0.0`
- `rlp` bumped to `>=3.0.0,<4.0.0`

### Notes

Having to do this is a consequence of a bigger problem, namely using upper bounds in dependencies. I suspect this was initially caused by bringing development practices from JS into Python. In JS this works because the every dependency has its own subdependencies which do not intersect; in Python the dependency namespace is flat, and upper bounds cause more problems than they solve (the fact that I only needed to bump the bounds without changing any code illustrates that). In this PR I tried to follow the general scheme of `eth-*` packages to set the new upper bounds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ extras_require = {
     'py-evm': [
         # Pin py-evm to exact version, until it leaves alpha.
         # EVM is very high velocity and might change API at each alpha.
-        "py-evm==0.5.0a3",
+        "py-evm==0.6.0a1",
         "eth-hash[pysha3]>=0.1.4,<1.0.0;implementation_name=='cpython'",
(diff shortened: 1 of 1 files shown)


fjarri:
Closed in favor of #238 - sorry, didn't notice at first.

PR #244, opened 2022-09-28
Last bit of changes to accommodate eth-abi ``4.0.0-beta.1``

fselmo:
### What was wrong?

Some refactoring was still necessary to accommodate `eth-abi` version `4.0.0-beta.1` which includes a correction in abi zero-padding.

### How was it fixed?

- `decode_single()` and `decode_abi()` -> `abi.encode()`
- Refactor of `from eth_abi import encode`, etc..., to `from eth_abi import abi` and using `abi.encode()` for more clarity. This just reduces cognitive overhead for a reader, especially with common words like "encode" and "decode" to know that it's specifically `abi.encode()` and `abi.decode()`.

### Miscellaneous updates
  - Updated `towncrier` types to match `ethereum-python-project-template` repo and to be able to use the `internal` type.
  - Added the commit hash for the `black` linting commit to a `.git-blame-ignore-revs` file to be able to optionally remove that blame history locally.

### To-Do:


pacrob (approved):
lgtm!

hamzawajahat (commented):
6,7

PR #245, opened 2022-09-30
Add support for ``ParisVM`` "the merge"

fselmo:
### What was wrong?

- ``ParisVM`` / merge support is being added to ``py-evm``. Make relevant changes to get `eth-tester` working with PoS here.

### How was it fixed?

- Needs ethereum/py-evm#2080 to be merged
- Add support for `safe` and `finalized` block identifiers
- Add support for generating correct genesis headers with post-merge / EIP-3675 rules (i.e. `difficulty` always 0, `nonce` always zero value...)

### To-Do:

- [x] Re-think block building, setting a random value for `mix_hash` each time
- [x] ~~Refactor some of the nomenclature around "mining" to be more general~~ -- separate PR perhaps? [yes, separate PR]


pacrob (approved):
lgtm!

eth_tester/backends/pyevm/main.py
```diff
@@ -128,16 +129,24 @@ def generate_genesis_state_for_keys(account_keys, overrides=None):
 
 
 def get_default_genesis_params(overrides=None):
-    # commented out params became un-configurable in py-evm during London refactor
+    # Commented out params became un-configurable in py-evm during London refactor.
+    # Post-merge now. Set genesis params to expect post-merge validation. If PoS field
+    # value defaults are not desired, use the overrides option.
+    from eth.constants import (
```
kclowes, line 135: I'm curious why you put these here instead of at the top of the file?

eth_tester/backends/base.py
```diff
@@ -31,7 +33,7 @@ def time_travel(self, to_timestamp):
     # Mining
     #
     @abstractmethod
-    def mine_blocks(self, num_blocks=1, coinbase=None):
+    def mine_blocks(self, num_blocks=1, coinbase=ZERO_ADDRESS):
```
kclowes, line 36: Why did we need to change this from `None` -> `ZERO_ADDRESS`?

eth_tester/backends/pyevm/main.py
```diff
@@ -419,15 +437,20 @@ def time_travel(self, to_timestamp):
...
+    # Importing blocks
     #
     @to_tuple
-    def mine_blocks(self, num_blocks=1, coinbase=None):
-        if coinbase is not None:
-            mine_kwargs = {"coinbase": coinbase}
-        else:
-            mine_kwargs = {}
+    def mine_blocks(self, num_blocks=1, coinbase=ZERO_ADDRESS):
+        from eth.vm.forks.paris import ParisVM
```
kclowes, line 444: Also curious about this import being here instead of at the top of the file?

kclowes (approved):
Woohoo! I had a couple questions for my own knowledge/understanging, but nothing that should block merging!

eth_tester/backends/pyevm/main.py
fselmo: I went back and forth on it a bit but I think up top may be better. We do have to play the game of `is_supported_pyevm_version_available()` else `None`. I think that's ok since these methods should never really be called if py-evm isn't available. Thoughts?

eth_tester/backends/base.py
fselmo: I actually added a `coinbase` that should pass an address validation now. We can definitely change this to optional if we want but this just changes the default address to the zero address and passes the address validation. I could go either way here.

fselmo: On second thought, let me double check this... the validation was already there. I think this broke something early on.

fselmo: Ok yeah... either [this logic](https://github.com/ethereum/eth-tester/pull/245/files#diff-184f0343bc8af5da7a16faac90ae2ecde91970b2824e75699d9dd86fb69faf90L353-L358) needs to be repeated everywhere (e.g. `mine_block`... not a good solution), or we add the `None` check to the actual inbound validation method for this field, or we set it to the zero address and leave the validation alone. Can`coinbase` be empty on mainnet?

eth_tester/backends/pyevm/main.py
kclowes: Yeah, I think up top might be a little clearer, but not a strong preference

eth_tester/backends/base.py
fselmo: @kclowes looks like we set `coinbase` to the zero address in py-evm as well, so atm coinbase is almost always the zero address unless it's passed in. I'm ok with either direction here. Let me know if you prefer one direction over another.

kclowes: Okay, I think it makes sense to leave it as the `ZERO_ADDRESS` then. Thanks for looking!

fjarri:
Sorry for raising a merged PR, but it seems like the best place to ask since the change was introduced here. I tried to look it up, but Ethereum RPC standard is not very rigorously documented. 

I am wondering, what prompted the change from `miner` to `coinbase` in the returned block data? When I query a block from Infura, or from a Geth node, it still has `miner` in it, and no `coinbase`. Is it implementing some standard that hasn't been introduced to mainnet yet?

PR #249, opened 2022-11-14
Fix ghost blocks

fselmo:
### What was wrong?

`get_block` methods were able to retrieve the pending block by number before it was added to the chain. This is misleading since the block has not yet been added to the chain. The pending block should only be retrieved via the block identifier `pending`.

### How was it fixed?

- Change logic to only allow the pending block to be retrieved via the `pending` block identifier for `get_block` methods.
- Test for the above changes

### To-Do:


kclowes (approved):
LGTM! 🚢

pacrob (approved):
:ship:

PR #261, opened 2023-06-22
Better UX for contract deployment

fselmo:
### What was wrong?

Don't return a contract address if the contract deployment transaction failed. This can only lead to issues.

### To-Do:


kclowes (approved):
:shipit:

PR #267, opened 2023-07-14
Configurable HDPath for Mnemonic

### LINKED ISSUE #259: Support HDPath for Mnemonic
* Version: 0.9.0b1
* Python: 3.10
* OS: linux

### What was wrong?

Currently when initializing the PyEVMBackend from a mnemonic, there is no option pass in an HDPath, which then uses the default HDPath which is incompatible with some other types of providers (such as Anvil)

https://github.com/ethereum/eth-tester/blob/master/eth_tester/backends/pyevm/main.py#L138

### How can it be fixed?

Allow passing in an HDPath to override the default

reedsa:
### What was wrong?

closes #259

### How was it fixed?

Parameterize the mnemonic hd_path for the pyevm backend.

### To-Do:

#### Cute Animal Picture

Cute animal picture


eth_tester/backends/pyevm/main.py
```diff
@@ -131,12 +131,12 @@ def get_default_account_keys(quantity=None):
...
 
 @to_tuple
-def get_account_keys_from_mnemonic(mnemonic, quantity=None):
+def get_account_keys_from_mnemonic(mnemonic, quantity=None, hd_path_override=None):
     keys = KeyAPI()
     seed = seed_from_mnemonic(mnemonic, "")
     quantity = quantity or 10
     for i in range(0, quantity):
-        hd_path = HDPath(f"m/44'/60'/0'/{i}")
+        hd_path = HDPath(f"{hd_path_override}'/{i}") or HDPath(f"m/44'/60'/0'/{i}")
```
reedsa, line 139: In testing, I needed a unique HDPath for each to generate all the expected accounts. We probably want to allow passing one hd_path for each account they expect to create. Curious of thoughts on this.

pacrob (approved):
lgtm!

eth_tester/backends/pyevm/main.py
```diff
@@ -131,12 +131,15 @@ def get_default_account_keys(quantity=None):
...
 
 @to_tuple
-def get_account_keys_from_mnemonic(mnemonic, quantity=None):
+def get_account_keys_from_mnemonic(mnemonic, quantity=None, hd_path_override=None):
     keys = KeyAPI()
     seed = seed_from_mnemonic(mnemonic, "")
     quantity = quantity or 10
     for i in range(0, quantity):
         hd_path = HDPath(f"m/44'/60'/0'/{i}")
+        if hd_path_override is not None:
```
fselmo, line 140: nit: I'm seeing we use `hd_path` as the kwarg in all other methods. We could use that here as well (instead of `hd_path_override`) and, if it's `None`, set the internal value, also so we don't set the `hd_path` variable twice with each iteration.

tests/backends/test_pyevm.py
```diff
@@ -261,6 +261,37 @@ def test_from_mnemonic(self):
...
+        balance = to_wei(15, "ether")  # Give each account 15 Eth
+        pyevm_backend = PyEVMBackend.from_mnemonic(
+            MNEMONIC,
+            num_accounts=num_accounts,
+            genesis_state_overrides={"balance": balance},
+            hd_path="m/44'/60'/0'/7",
+        )
+
+        # Each of these accounts stems from the MNEMONIC
+        expected_accounts = [
```
fselmo, line 277: Since I don't really know what public addresses are actually expected with the provided parameters, I'm not very sure what this is testing as someone just reading the code other than that it doesn't break anything.

Maybe we could test here that the `hd_path` at least has some influence on the account generation by comparing with / without the `hd_path` kwarg. Or is there another good marker that we're testing already here? I haven't dived too deep.

fselmo: To be fair, the above note goes for the mnemonic test above this one too. It would be nice if that also checked that it at least influences the addresses generated. Maybe there's a good fix here for both if we abstract the default expected accounts into a variable at the top `DEFAULT_EXPECTED_ACCOUNTS = [...]` and add one-liners to each of these: `assert not any(acct in DEFAULT_EXPECTED_ACCOUNTS for acct in tester.get_accounts()`.

reedsa: For testing the hd_path, I've added a couple of those assertions to illustrate the accounts are different than those created with a mnemonic alone. I've left the original checks for account balances in the test above, I dont think that is covered anywhere else, but added the assertion for the default mnemonic accounts there as well.

fselmo: Oh yeah I think that works well too since `MNEMONIC` is a constant 👌🏼

fselmo (approved):
lgtm 👍🏼 👍🏼

PR #269, opened 2023-07-31
fee history endpoint

### LINKED ISSUE #258: Feature request: eth_feeHistory RPC endpoint
* Version: 0.9.0b1
* Python: 3.11
* OS: linux

### What was wrong?

As of today, we can't call the `eth_feeHistory` RPC endoint, it will end on:

```python
ValueError: RPC Endpoint has not been implemented: eth_feeHistory
```

### How can it be fixed?

Inspiration may be taken from https://github.com/ethereum/web3.py/pull/2094.

reedsa:
### What was wrong?
Closes #258

### How was it fixed?

Add RPC endpoint to backends for eth_feeHistory

### To-Do:

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,6 +594,14 @@ value from the evm.
 Executes the provided `transaction` object, measuring and returning the gas
 consumption.
 
+<a id="api-fee_history"></a>
+
+#### `EthereumTester.get_fee_history(block_count=1, newest_block='latest', reward_percentiles=[])`
+
+Return the historical gas information for the number of blocks specified as the `block_count` starting from `newest_block`.
+Note that specifying `reward_percentiles` has no effect on the response and so `reward` will always return an  empty list.
+
+
 
 ### Logs and Filters
 
--- a/README.md
+++ b/README.md
@@ -599,7 +599,7 @@ consumption.
 #### `EthereumTester.get_fee_history(block_count=1, newest_block='latest', reward_percentiles=[])`
 
 Return the historical gas information for the number of blocks specified as the `block_count` starting from `newest_block`.
-Note that specifying `reward_percentiles` has no effect on the response and so `reward` will always return an  empty list.
+Note that specifying `reward_percentiles` has no effect on the response and so `reward` will always return an empty list.
 
 
 
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -524,6 +524,37 @@ class PyEVMBackend(BaseChainBackend):
             is_pending,
         )
 
+    def get_fee_history(
+        self, block_count=1, newest_block="latest", reward_percentiles: List[int] = []
+    ):
+        if isinstance(block_count, int) and not 1 <= block_count <= 1024:
+            raise ValidationError("block_count must be between 1 and 1024")
+
+        if newest_block == "pending":
+            raise NotImplementedError(
+                '"pending" block identifier is unsupported in eth-tester'
+            )
+
+        block = self.get_block_by_number(newest_block)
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -594,6 +594,14 @@ value from the evm.
 Executes the provided `transaction` object, measuring and returning the gas
 consumption.
 
+<a id="api-fee_history"></a>
+
+#### `EthereumTester.get_fee_history(block_count=1, newest_block='latest', reward_percentiles=[])`
+
+Return the historical gas information for the number of blocks specified as the `block_count` starting from `newest_block`.
+Note that specifying `reward_percentiles` has no effect on the response and so `reward` will always return an  empty list.
```
fselmo, line 602: nit, extra space

```suggestion
Note that specifying `reward_percentiles` has no effect on the response and so `reward` will always return an empty list.
```

eth_tester/backends/pyevm/main.py
```diff
@@ -524,6 +524,37 @@ def get_transaction_receipt(self, transaction_hash):
             is_pending,
         )
 
+    def get_fee_history(
+        self, block_count=1, newest_block="latest", reward_percentiles: List[int] = []
+    ):
+        if isinstance(block_count, int) and not 1 <= block_count <= 1024:
+            raise ValidationError("block_count must be between 1 and 1024")
+
+        if newest_block == "pending":
```
fselmo, line 533: Hmm eth-tester does support the "pending" block identifier. I do see this message in the `estimate_gas()` method but I'm not sure what that means exactly.

I just tried `eth_feeHistory` against mainnet with "pending" and it is a valid identifier for the endpoint. We should probably get rid of this validation here and just set `newest_block` to "latest" if "pending" is passed in. I think that makes sense from an API standpoint too since that's a valid block id but the header isn't a proper header yet... so we just return the latest possible block if "pending" is used.

Gut check / thoughts on that idea?

Honestly seems like it's a bug in `estimate_gas()` and we should handle this case similarly there but that's unrelated.

eth_tester/utils/backend_testing.py
```diff
@@ -238,6 +238,105 @@ def test_get_nonce(self, eth_tester):
...
+                1,
+                "earliest",
+                [],
+                {
+                    "base_fee_per_gas": [],
+                    "gas_used_ratio": [],
+                    "reward": [],
+                },
+            ],
+        ],
```
fselmo, line 287: If we agree to add "pending" as a valid API value as per the comment above, we should add the test case here.

newsfragments/258.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for ``eth_getFeeHistory`` for ``PyEVMBackend`` via ``get_fee_history()`` method.
```
fselmo, line 1:
```suggestion
Add support for ``eth_feeHistory`` for ``PyEVMBackend`` via ``get_fee_history()`` method.
```

fselmo (commented):
Looks good. I like this simple approach. I have a question on the handling of the "pending" block_id. I think anywhere the block identifier is available as an argument, all block Ids are valid. I think the logic might change depending on the context so, in this case, using the "latest" block for "pending" I think makes good sense but curious on your thoughts.

eth_tester/backends/pyevm/main.py
reedsa: Glad to hear I'm not the only one confused by that check. I think I had an issue with the pending block in one of the tests but I'll apply the logic to just go with "latest" and that will take care of it!

fselmo (approved):
lgtm 👌🏼

PR #281, opened 2024-01-25
bugfix: properly convert access list storage keys to int base 16

fselmo:
### What was wrong?

- Lacking proper tests, we were not converting access list storage keys to int with base 16.

### How was it fixed?

- `int(value)` -> `int(value, 16)`
- Add tests

### Todo:

- [x] Clean up commit history


kclowes (approved):
LGTM! 🚀

pacrob (approved):
lgtm!

PR #285, opened 2024-04-03
Additional changes to fully support Cancun

fselmo:
- Pull out the blob data when sending a transaction to the execution layer (py-evm). The execution layer does not deal with blob data, that's the whole point of blob transactions. We pull out the data and re-send with only the versioned hashes.

- Sum up the correct value for `blob_gas_used` during block proposal. This is validated in the execution layer (py-evm).

---

**Note: this will require: ethereum/py-evm#2165  and ethereum/py-evm#2166 to be merged and released. I've tested locally and these changes make it so we can send a blob transaction via *web3.py* -> `EthereumTesterProvider` -> *eth-tester*  -> *py-evm*.**

### How was it fixed?

- Remove the blob data and re-serialize blob transactions in a way that the EL expects, without the bloat of the blob data.
- Sum up `blob_gas_used` and send it to EL (py-evm) 
- Add validation and normalization for Cancun-related fields on block / header and transactions.

### Todo:

- [x] Clean up commit history
- [x] More testing + verify these are all of the remaining changes related to full Cancun support

#### Cute Animal Picture

Screenshot 2024-04-03 at 17 54 20


eth_tester/normalization/common.py
```diff
@@ -2,7 +2,7 @@
     encode_hex,
     is_hex,
     to_dict,
-    to_tuple,
+    to_list,
```
kclowes, line 5: curious about this change?

eth_tester/backends/mock/serializers.py
```diff
@@ -52,6 +52,13 @@ def serialize_full_transaction(transaction, block, transaction_index, is_pending
         partial(assoc, key="type", value=extract_transaction_type(transaction)),
     )
 
+    if int(serialized_transaction["type"], 16) > 0:
+        serialized_transaction = assoc(
+            serialized_transaction,
```
fselmo, line 63: This was a bonus that I noticed clients return. There is still a `v` field for non-legacy transactions but there's also the correct `y_parity` field of the signature.

eth_tester/backends/pyevm/main.py
```diff
@@ -734,7 +750,24 @@ def _create_type_aware_unsigned_transaction(self, normalized_txn):
 
     def send_raw_transaction(self, raw_transaction):
         vm = _get_vm_for_block_number(self.chain, "latest")
-        evm_transaction = vm.get_transaction_builder().decode(raw_transaction)
+
+        if raw_transaction[0] == 3:
```
reedsa: Nit: use a constant to make it easy to tell what the number 3 means.

eth_tester/backends/pyevm/serializers.py
```diff
@@ -74,13 +77,30 @@ def serialize_block(block, full_transaction, is_pending):
...
         base_fee = block.header.base_fee_per_gas
         block_info.update({"base_fee_per_gas": base_fee})
 
-    if hasattr(block.header, "withdrawals_root") and hasattr(block, "withdrawals"):
+    if hasattr(block.header, "withdrawals_root"):
+        # shanghai
         block_info.update({"withdrawals": serialize_block_withdrawals(block)})
         block_info.update({"withdrawals_root": block.header.withdrawals_root})
 
+    if all(
```
reedsa: Similar to other comment, I wonder if this check is useful elsewhere. If so it might be worth calling a boolean function with a descriptive name.

Otherwise I'm good with the comment you used in this block.

tests/backends/test_pyevm.py
```diff
@@ -432,3 +443,54 @@ def test_eth_get_storage_at(self):
             assert tester.get_storage_at(acct, HexStr("0x0")) == f"0x{'00'*32}"
             assert tester.get_storage_at(acct, HexStr("0x1")) == f"0x{'00'*31}01"
             assert tester.get_storage_at(acct, HexStr("0x2")) == f"0x{'00'*31}02"
+
+    # --- cancun network upgrade --- #
+
+    BLOB_TEXT = "We are the music makers, And we are the dreamers of dreams."
```
reedsa, line 449: :smile:

eth_tester/backends/pyevm/serializers.py
```diff
@@ -103,13 +123,8 @@ def serialize_transaction(block, transaction, transaction_index, is_pending):
...
         "gas": transaction.gas,
         "data": transaction.data,
-        "r": transaction.r,
-        "s": transaction.s,
-        "v": transaction.v
-        if _field_in_transaction(transaction, "v")
-        else transaction.y_parity,
     }
-    if _field_in_transaction(transaction, "gas_price"):
+    if int(txn_type, 16) in (0, 1):
```
reedsa: Moar nit: Worth a constant here as well? As a naive person reading I don't know offhand what `txn_type` being 0 or 1 means.

Comment or descriptive name for this could help.

reedsa (approved):
Nice work, I left a couple comments you can take or leave. Not a blocker in any sense.

eth_tester/validation/outbound.py
```diff
@@ -277,6 +312,19 @@ def _validate_fork_specific_fields(block):
         partial(validate_array, validator=validate_withdrawal)(block["withdrawals"])
         validate_32_byte_string(block["withdrawals_root"])
 
+    # Cancun fork
+    if all(
+        _ not in block
```
kclowes: I like @reedsa's suggestion to make these into methods. I think it would be easier to reason about.

kclowes (approved):
Looks good to me! I put this in the comments too, but I like @reedsa's idea to have descriptive method names instead of `if any(...)` and `if all(...)`

eth_tester/normalization/common.py
fselmo: Consistent with output from clients. They are usually lists, not tuples in the JSON.

pacrob (approved):
Looks good! I like the Stu idea of using constants for transaction types for readability. Whaddya think?

fselmo:
@reedsa @pacrob @kclowes the [last commit](https://github.com/ethereum/eth-tester/pull/285/commits/448c9da14b6ebc51332a74e67b98ed7d9c628be6) should have some refactors that I think are along the lines of what was asked for. Will wait for extra eyes there before merging.

reedsa (approved):
Looks awesome, great refactor!

PR #287, opened 2024-04-08
eth-account bugfix to remove pydantic dependency; might as well use it

fselmo:
### What was wrong?

This is technically not necessary but since *eth-account* had a dependency bugfix let's avoid any sort of issues altogether by lower-pinning cancun support for eth-account beyond the bugfix.

### How was it fixed?

- use `eth-account>=0.11.2` (bugfix patch release that removes the dependency on pydantic introduced in `0.11.1`)

### Todo:

- [x] Clean up commit history

### DIFF
--- a/newsfragments/287.internal.rst
+++ b/newsfragments/287.internal.rst
@@ -1,0 +1,1 @@
+Bump `eth-account` to ``>=0.11.2`` since there was a dependency-related bugfix.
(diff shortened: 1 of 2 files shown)


reedsa (approved):
lgtm

pacrob (approved):
lgtm!

PR #288, opened 2024-04-15
feat: remove setuptools dependency

wakamex:
### What was wrong?

using pkg_resources which comes from setuptools disallows me from installing ape with `uv` without adding setuptools as a regular dependency to ape.
working toward fully removed dependence on setuptools in ape, tracking PR here: https://github.com/ApeWorX/ape/pull/2003
also, moving away from pkg_resources is best practice. its features are now available in the standard library.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_tester/backends/pyevm/utils.py
+++ b/eth_tester/backends/pyevm/utils.py
@@ -2,14 +2,19 @@ from __future__ import (
     absolute_import,
 )
 
+from importlib.metadata import (
+    PackageNotFoundError,
+    version,
+)
 from typing import (
     Any,
     Dict,
     Union,
 )
 
-from importlib.metadata import version, PackageNotFoundError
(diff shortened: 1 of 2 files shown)


wakamex:
should I be manually changing CHANGELOG.rst?

wakamex:
I read the "How to bumpversion" in the README and my interpretation is the version bump happens after a PR gets merged into main, not before or during (due to "This is typically done from the main branch")

pacrob:
> version bump happens after a PR gets merged into main

That's true, it'll happen with the next release. With each PR, we want a newsfragment file like the one I've added. Upon release, those get compiled into the changelog.

Thanks for the update! I think (hope) that this is the last instance of `pkg_resources` we'll have to deal with.

Issue #211, opened 2021-11-03
Eth-tester cannot validate signed calls from `web3.py`


Pet3ris:
Hi There,

I was pointed to the `eth-tester` repo with this issue: https://github.com/ethereum/web3.py/issues/2198.

It seems like there is an issue with validating signed calls when using eth-tester as a provider?

PR #297, opened 2024-09-16
bugfix: `eth_call` for pyevm should not check for known account 

### LINKED ISSUE #2198: Cannot use contract calls with signing middleware: invalid "from" key error
* Version: latest web3["tester"]
* Python: 3.8
* OS: osx

### What was wrong?

```python
MOCK_CONTRACT = """
val: public(uint256)
"""

MOCK_ABI = """[
    {
        "inputs": [],
        "name": "val",
        "outputs": [
            {
                "internalType": "uint256",
                "name": "",
                "type": "uint256"
            }
        ],
        "stateMutability": "view",
        "type": "function"
    }
]
"""

def deploy_contract(w3):
    compiled = compile_code(MOCK_CONTRACT)
    deployment_bytecode = compiled["bytecode"]
    vault = w3.eth.contract(
        abi=MOCK_ABI, bytecode=deployment_bytecode
    )

    # Deploy vault contract
    receipt = w3.eth.wait_for_transaction_receipt(
        vault.constructor().transact()
    )
    return receipt["contractAddress"]

def test_call_signed():
    w3 = Web3(EthereumTesterProvider())
    acct = Account.create("TEST ACCOUNT")
    w3.eth.send_transaction({"to": acct.address, "value": 10**9})
    w3.middleware_onion.add(construct_sign_and_send_raw_middleware(acct))
    w3.eth.default_account = acct.address

    contract_address = deploy_contract(w3)
    contract = w3.eth.contract(address=contract_address, abi=MOCK_ABI)
    _ = contract.functions.val().call()
```

This test fails with the error `No valid "from" key was provided`. Please note that the deployment transaction succeeds but the contract call fails even though contract calls don't require signing.

I've also confirmed that the calls work well without the signing middleware.

### How can it be fixed?

Could we remove the from check from `call()` altogether even in signed middleware mode?

fselmo:
### What was wrong?

closes https://github.com/ethereum/web3.py/issues/2198

### How was it fixed?

- Build an unsigned, spoofed tx rather than a signed tx for `eth_call` via *py-evm*. The account should not need to be a known account since we aren't signing for `eth_call` because it does not change state.

### Todo:

- [x] Clean up commit history

### DIFF
--- a/eth_tester/backends/pyevm/main.py
+++ b/eth_tester/backends/pyevm/main.py
@@ -873,14 +873,15 @@ class PyEVMBackend(BaseChainBackend):
         if "gas" not in defaulted_transaction:
             defaulted_transaction["gas"] = self._max_available_gas()
 
-        signed_evm_transaction = self._get_normalized_and_signed_evm_transaction(
+        unsigned_tx = self._get_normalized_and_unsigned_evm_transaction(
             defaulted_transaction,
             block_number,
         )
+        evm_transaction = EVMSpoofTransaction(unsigned_tx, from_=transaction["from"])
 
         computation = _execute_and_revert_transaction(
             self.chain,
-            signed_evm_transaction,
+            evm_transaction,
             block_number,
         )
         if computation.is_error:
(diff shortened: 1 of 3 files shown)


reedsa (approved):
lgtm

pacrob:
This: https://github.com/ethereum/eth-tester/issues/211

was opened in relation to the original `web3.py` issue. Can this close that too?

pacrob (approved):
Original user issue no longer appears. :sheep: it!

PR #304, opened 2024-11-12
refactor(`backends/mock/factory`): replace long if statements with dict

dzhalaevd:
### What was wrong?
The code had multiple long if-else statements, making it repetitive and hard to read.

### How was it fixed?
Replaced the if-else statements with a dictionary of default values. This simplifies the logic and improves readability.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


pacrob:
Hey @DavidRomanovizc, sorry for the delay in responding. We're actually going to dropping `MockBackend` in the next major release, so it doesn't make much sense to add optimizations for it now. Thank you for your contributions though!

PR #325, opened 2025-04-22
Normalize snake_case to camelCase keys

### LINKED ISSUE #324: eth-tester returns camelCase, not snake_case
Update serializers in py-evm and eels to send outbound payloads with camelCase keys

reedsa:
### What was wrong?
 
Closes #324

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_tester/normalization/common.py
+++ b/eth_tester/normalization/common.py
@@ -13,7 +13,7 @@ from eth_tester.utils.encoding import (
 )
 
 
-def to_lower_camel_case(value):
+def to_lower_camel_case(value: str) -> str:
     """
     Convert a string to lower camel case.
     """
--- a/eth_tester/normalization/common.py
+++ b/eth_tester/normalization/common.py
@@ -13,13 +13,37 @@ from eth_tester.utils.encoding import (
 )
 
 
-def to_lower_camel_case(key):
+def to_lower_camel_case(value):
     """
     Convert a string to lower camel case.
     """
     return "".join(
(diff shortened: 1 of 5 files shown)


reedsa:
Before updating tests I wanted to check my approach to make sure I'm heading in the right direction.

fselmo:
Yep. This is a great, simple way to handle it. We just need to make sure all dicts go through this method on the way out but I think this is the case already.

reedsa:
@fselmo Many of them are hitting the `identity` and I think I need to change those to format dict keys in order for this to work. I'll change that and see what else I find.

fselmo:
> @fselmo Many of them are hitting the `identity` and I think I need to change those to format dict keys in order for this to work. I'll change that and see what else I find.

Ah, right. Well, we could move all objects through a formatter that recursively formats keys for all dicts, even within lists, etc, if that's easier? Don't think that would take too much of a hit in performance.

So have most "normalizers" be for normalizing the values and then have a key normalizer that all objects go through that recursively camels <-> snakes depending on inbound or outbound.

Whatever makes more sense as you're in there. Just figured I'd throw that out there.

---
edit: Actually I think serializers will handle the outbound so nm. Whatever works best here.

reedsa:
Waited too long to rebase against `next`. Created a new branch and cherry-picked these commits, new PR is here: https://github.com/ethereum/eth-tester/pull/330

PR #326, opened 2025-04-22
feature: EELS backend

fselmo:
### What was wrong?

- It would be a welcomed feature to be able to communicate with EELS as the backend. In the future, if a thin server can be wrapped around *eth-tester*, making calls to EELS could be used to test EIPs or any call against the official execution layer specifications.

### How was it fixed?

- This PR scrapes together an implementation for talking to EELS. It takes into account all type conversions and is working against the [entire web3.py test suite](https://github.com/ethereum/web3.py/pull/3447). What it provably isn't is an efficient implementation. There are likely performance improvements that can be made on top of this initial implementation, as is evident with the core tests in web3.py taking ~30min to complete with the EELS backend.
- Seeing that it's passing web3.py tests, and this is going into a feature branch, I would advocate for adding tests in a separate PR but could be convinced otherwise.


pacrob (approved):
merge it!

PR #328, opened 2025-04-24
Drop mockbackend

### LINKED ISSUE #327: Remove `MockBackend` in next major release
While we work on `next` branch, let's rip out the unmaintained `MockBackend`. With a new useful `EELSBackend` coming in, the mocked hard-coded backend is a chore to maintain and it honestly feels rather useless.

pacrob:
### What was wrong?

`MockBackend` is not useful, getting rid of it.

Closes #327 

### How was it fixed?

 - Removed all functionality related to `MockBackend`
 - Added a couple functions related to checking if `EELSBackend` is available

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


kclowes (commented):
-958, nice! I think the eels stuff should be pulled into a separate PR for when we add it in since it looks like it might not really work right now, but what do you think about that?

eth_tester/backends/__init__.py
```diff
@@ -10,36 +14,52 @@
     EELSBackend,
     eels_is_available,
 )
-from .mock import (
-    MockBackend,
-)
 from .pyevm import (
     PyEVMBackend,
+    is_supported_eels_version_available,
```
fselmo: why are we importing `is_supported_eels_version` from pyevm? This is what `eels_is_available` is for, I believe, no?

eth_tester/backends/pyevm/utils.py
```diff
@@ -27,11 +27,31 @@ def get_pyevm_version():
         return None
 
 
+def get_eels_version():
```
fselmo: This shouldn't belong in pyevm backend. I believe there is a `eels_is_available()` that should help here.

pacrob:
> -958, nice! I think the eels stuff should be pulled into a separate PR for when we add it in since it looks like it might not really work right now, but what do you think about that?

Since the related backend and `is_available` function were already there, I just factored them in. Fine to leave for another PR if you prefer.

fselmo:
> I think the eels stuff should be pulled into a separate PR for when we add it in

@kclowes what do you mean by "when we add it in" here?

fselmo:
> -958, nice!

💯

eth_tester/backends/eels/utils.py
```diff
@@ -1,12 +1,36 @@
...
     try:
-        import ethereum  # noqa: F401
+        import ethereum_execution  # noqa: F401
 
         return True
     except ImportError:
         return False
 
 
+def get_eels_version():
```
fselmo: I don't think we need this. I'm not a big fan of the old approach to verify the version in the code. I think we should control versioning in the dependencies and leave it be. Let's just use `eels_is_available()`.

Thoughts?

pacrob: no complaints

fselmo: We should probably drop the version check in the code for py-evm as well and control that in the versioning pins as we already do. That seems like a stale check. Can be done in this PR or another.

pacrob: I'll let this be just drop MockBackend, I think

fselmo: sgtm 👍🏼

fselmo (approved):
lgtm! I'd like to see if we resolved @kclowes' comments before merging. 

Also, I don't think we should leave the state of the backends both as install extras with conditionals. We talked about perhaps defaulting to EELS and we should remove the EELS conditionals and bake it into the requirements if so. Then we can leave the py-evm conditionals as they were before and just remove the `..._version_is_available` stale check. Issue created to track that here: #329

kclowes:
> @kclowes what do you mean by "when we add it in" here?

I don't see the eels dependency/extra anywhere yet, so it feels to me like it's sort of half-baked to be checking for it, even if it might be installed in the local environment. Alternatively, if it's just adding the extra dependency, I'd be fine adding that in now, but it seems like there is a logical separation that might be clearer as a separate PR. I'd rather keep this PR to just removing the MockBackend, but don't have super strong feelings on that if it's easier to keep those methods and add the eels extra now.

fselmo:
> I don't see the eels dependency/extra anywhere yet, so it feels to me like it's sort of half-baked to be checking for it, even if it might be installed in the local environment.

[This](https://github.com/ethereum/eth-tester/blob/next/setup.py#L32-L34)?

Agreed, I think we can tackle the default option in #329

kclowes:
lol, I'm blind. I was looking at main. carry on 😄

kclowes (approved):

PR #331, opened 2025-04-29
Return camelCase keys from serializers

### LINKED ISSUE #321: eth-tester serialization on the way out
Each back end has a serializer to convert values from the client. Both py-evm and eels use snake_case, so we'll need to convert snake_case values to camelCase. Will need to reference the [JSON-RPC specs](https://ethereum.github.io/execution-apis/api-documentation/) to get the right values here.

### LINKED ISSUE #322: eth-tester normalize like a client on the way out
reference [JSON-RPC spec](https://ethereum.github.io/execution-apis/api-documentation/) to make sure correct types are returned.

reedsa:
### What was wrong?

Closes #321
Closes #322

### How was it fixed?
Outbound serializers, validators and normalizers have been updated to expect all dicts to have camel case keys.

### Todo:

- [X] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -305,18 +305,7 @@ jobs:
         environment:
           TOXENV: py313-pyevm
 
-  py38-eels:
-    <<: *common
-    docker:
-      - image: cimg/python:3.8
-        environment:
-          TOXENV: py38-eels
-  py39-eels:
-    <<: *common
-    docker:
-      - image: cimg/python:3.9
-        environment:
-          TOXENV: py39-eels
+  # eels only supported on Python 3.10+
   py310-eels:
     <<: *common
     docker:
@@ -371,8 +360,6 @@ define: &all_jobs
   - py311-pyevm
   - py312-pyevm
   - py313-pyevm
-  - py38-eels
-  - py39-eels
   - py310-eels
   - py311-eels
   - py312-eels
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -305,6 +305,43 @@ jobs:
         environment:
           TOXENV: py313-pyevm
 
+  py38-eels:
+    <<: *common
+    docker:
+      - image: cimg/python:3.8
+        environment:
+          TOXENV: py38-eels
+  py39-eels:
+    <<: *common
+    docker:
+      - image: cimg/python:3.9
+        environment:
+          TOXENV: py39-eels
+  py310-eels:
+    <<: *common
+    docker:
+      - image: cimg/python:3.10
+        environment:
+          TOXENV: py310-eels
+  py311-eels:
+    <<: *common
+    docker:
+      - image: cimg/python:3.11
+        environment:
+          TOXENV: py311-eels
+  py312-eels:
+    <<: *common
+    docker:
+      - image: cimg/python:3.12
+        environment:
+          TOXENV: py312-eels
+  py313-eels:
+    <<: *common
+    docker:
+      - image: cimg/python:3.13
+        environment:
+          TOXENV: py313-eels
+
 define: &all_jobs
   - docs
   - py38-core
@@ -334,6 +371,12 @@ define: &all_jobs
   - py311-pyevm
   - py312-pyevm
   - py313-pyevm
+  - py38-eels
+  - py39-eels
+  - py310-eels
+  - py311-eels
+  - py312-eels
+  - py313-eels
 
 workflows:
   version: 2
--- a/tests/backends/eels/test_eels_serializers.py
+++ b/tests/backends/eels/test_eels_serializers.py
@@ -1,3 +1,4 @@
+import pytest
 from typing import (
     Any,
     Dict,
@@ -9,6 +10,7 @@ from typing import (
 from ethereum.cancun.blocks import (
     Block,
     Log,
+    Withdrawal,
 )
 from ethereum.cancun.transactions import (
     LegacyTransaction,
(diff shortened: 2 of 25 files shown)


reedsa:
@fselmo I have the serializers doing the camelCasing for all keys now. The tests are still a WIP but worth a look if you have time. I still need to work on the inbound normalizers so it expects camel case keys as well.

.circleci/config.yml
```diff
@@ -334,6 +371,12 @@ define: &all_jobs
   - py311-pyevm
   - py312-pyevm
   - py313-pyevm
+  - py38-eels
```
kclowes: It doesn't look like the eels tests are actually running. is this intentional while we're in the inbetween here?

tests/backends/eels/test_eels_serializers.py
```diff
@@ -0,0 +1,215 @@
...
+def test_serialize_block_returns_camel_case_keys(
+    block: Union[Block, Dict[str, Any]],
+) -> None:
+    is_pending = isinstance(block, dict)
+    if isinstance(block, Block):
+        # Finalized blocks
+        num_transactions = len(block.transactions)
+        num_withdrawals = len(block.withdrawals)
+        transaction = block.transactions[0] if num_transactions > 0 else None
+    elif isinstance(block, dict):
```
kclowes: nit, could use `is_pending` here

tests/core/normalization/test_default_outbound_normalization.py
```diff
@@ -1,11 +1,129 @@
...
+        (f"0x{'01' * 20}", f"0x{'01' * 20}"),
+        (b"\x01" * 32, None),
+        (b"\x01" * 21, None),
+        (b"\x01" * 19, None),
+        (b"\x01" * 0, None),
+        (b"\x01" * 1, None),
+    ),
+)
+def test_outbound_account_normalization(value: Any, expected: Any) -> None:
+    if expected:
```
kclowes: We've been moving away from tests that have an if statement in them because it allows the test naming around the expectation to be clearer. What do you think about splitting this into 2 tests - maybe something like `test_valid_outbound_account_normalization` and `test_invalid_outbound...`. Feel free to take or leave!

tests/core/validation/test_outbound_validation.py
```diff
@@ -14,307 +16,452 @@
...
                 access_list=(("", (1, 2)),),
             ),
             False,
+            id="invalid_empty_string_address_in_blob_access_list",
+        ),
+        pytest.param({}, False, id="invalid_empty_dict"),
+        pytest.param(
+            merge(make_legacy_txn(), {"invalid-key": 1}),
+            False,
+            id="invalid_extra_key",
```
kclowes, line 358: Thank you for all of the ids here ⭐ that makes it much easier to see what each is testing!

kclowes (approved):
Looks good to me overall! I left a few comments, but nothing blocking. I think it would be good to make tests pass before merging, so maybe this actually becomes part of another feature branch on top of next where we can do the other normalization/validation too? If that seems too complicated, I'm okay merging into next knowing that test passing will be coming soon ™️

.circleci/config.yml
reedsa: Oh right, I meant to circle back to this. I am running into an issue with pip not finding the `ethereum-execution` package. Tried various tags of the package and ssh'ing to the ci box to try. Error is returned like the package does not exist.

```sh
pip install ethereum-execution
ERROR: Could not find a version that satisfies the requirement ethereum-execution (from versions: none)
ERROR: No matching distribution found for ethereum-execution
```

Any ideas what I might try next?

reedsa: Oh it seems the execution-specs only support python >3.10! I could remove the 3.8/3.9 versions of the eels jobs. I will isolate the dependency to only be part of those environments 🙏

reedsa: There we go, eels tests are now running! Also fixed lint. Now the failures are happening in tests that do internal API calls with camelCase payloads from the output of another call (Inbound calls still expect snake_case).

I am happy keeping this branch open until all conversions have been made. I can open new PRs against this branch and we can merge this uberbranch into next once things are stabilized!

reedsa:
@fselmo I am still a bit confused by transactions as hash values inside of a Block. When a block is serialized with `full_transactions=False` and I attempt to serialize with a list of txn hashes, a ValueError occurs. It appears that the backend returns the full transaction and then the serializer does the work to convert each to the respective hash for the block. The API spec defines either bytes or the full transaction so I expected that eels or py-evm would send back the hashes or full transaction based on the `full_transactions` request param.

I think I'm just confused where the responsibility lies in returning hashes.

tests/core/normalization/test_default_outbound_normalization.py
reedsa: Definitely didnt consider that, I'll remember to do that going forward. I'll update these and I also see some other tests that could be simpler if I split them into separate tests.

fselmo:
> I think I'm just confused where the responsibility lies in returning hashes.

@reedsa I think this may vary with each backend. The responsibility is ultimately on the client to _return_ the object as it was specified in the request. If the backend allows for this configuration, then we can pass the request param along to the backend, but if it doesn't, we have to provide it ourselves with *eth-tester* as the "client" (or simulation of a client) in this scenario.

tests/backends/eels/test_eels_serializers.py
reedsa: I chose to split these apart like I did for the "is_valid" or "expected" value tests, and perform the transaction checks in another test too.

fselmo (approved):
If this feels like it's in good shape for you to build on top of, this looks good to me to merge into feature branch 👍🏼

PR #334, opened 2025-05-08
Update Inbound validators, normalizers to use camelCase keys

### LINKED ISSUE #318: Normalize for the back end on the way in
py-evm/eels accept snake_case values, but values come in as camelCase.

reedsa:
### How was it fixed?
Inbound validators and normalizers have been updated to expect all dicts to have camel case keys.

### Todo:

- [X] Clean up commit history

#### Cute Animal Picture


tests/utils.py
```diff
@@ -36,6 +37,65 @@ def yield_key_value_if_value_not_none(
...
+    )
+    yield from yield_key_value_if_value_not_none("value", value)
+    yield from yield_key_value_if_value_not_none("data", data)
+    yield from yield_key_value_if_value_not_none("nonce", nonce)
+    yield from yield_key_value_if_value_not_none("accessList", access_list)
+    yield from yield_key_value_if_value_not_none("r", r)
+    yield from yield_key_value_if_value_not_none("s", s)
+    yield from yield_key_value_if_value_not_none("v", v)
+    yield from yield_key_value_if_value_not_none(
+        "blob_versioned_hashes", blob_versioned_hashes
```
fselmo: Looks like this is missing `camelCasing` 👀. @reedsa does that also mean we are missing good tests for this?

fselmo:
@reedsa I'm good to separate this into a separate PR but this does not validate like a client on the way in. We will need to accept integers as hex, etc, all of the expected values for JSON-RPC objects.

I would consider this PR just being the `snake_case` -> `camelCase` PR. It looks like github didn't pick up on the `Closes #...` line you put in the description anyway, but I would remove that from this. This would not close those tickets as we still need to:

- validate values like a client (refer to ethereum execution API spec for all these cases)
- normalize those new values for the backend (`to_integer_if_hex`, etc)

I think we can pause on the above two bullets until the pydantic research spike that I just created (#338) is done. But let's clarify in the description that this does not close these issues and we can more properly review it under a smaller scope of camel -> snake.

reedsa:
@fselmo I don't think I was clear in splitting out that new task. I had intended for this to be strictly just the camel casing and additional tests. The task I created was meant for validation of types. I can remove the closes line.

Were you thinking that task you just picked up was for something else?

fselmo:
> @fselmo I don't think I was clear in splitting out that new task. I had intended for this to be strictly just the camel casing and additional tests. The task I created was meant for validation of types. I can remove the closes line.
> 
> Were you thinking that task you just picked up was for something else?

I think we're on the same page. I created the pydantic issue as a sub-task under the one you created for validation of types. I think the only confusion came from the `Closes #...` line. Does that make sense?

tests/utils.py
reedsa: Good callout there, I see that this specific case is falling through because the "send" transaction type throws for any invalid key, but this should be failing for a different reason. I've added assertions on all exception messages to catch things like this going forward. Also fixed this camelCasing to make this fail as expected, since blobVersionedHashes are required to go through the raw transaction path.

eth_tester/backends/pyevm/main.py
```diff
@@ -627,21 +632,18 @@ def _normalize_transaction(self, transaction, block_number="latest"):
...
-            is_dynamic_fee_transaction or "access_list" in transaction
-        )
+        is_typed_transaction = is_dynamic_fee_transaction or "accessList" in transaction
 
         for key in transaction:
             if key in ("from", "type"):
                 continue
             if key == "v" and is_typed_transaction:
-                yield "y_parity", transaction[
+                yield "yParity", transaction[
```
fselmo, line 644: I can be convinced to ignore this here since we're going to experiment with the pydantic models, but part of normalization should be camelCase -> snake_case. Instead of "normalizing" and then calling `dict_keys_to_snake_case` randomly inside `_create_type_aware_unsigned_transaction` (e.g.), there should be a separation of logic where that doesn't need to be coerced into snake case outside of the normalization process.

Normalization is done to normalize the object for the backend. If the backend is expecting snake-case it should be handled in that step.

eth_tester/backends/pyevm/main.py
```diff
@@ -700,15 +702,18 @@ def _get_normalized_and_signed_evm_transaction(
         return evm_transaction.as_signed_transaction(signing_key)
 
     def _create_type_aware_unsigned_transaction(self, normalized_txn):
-        if all(_ in normalized_txn for _ in ("access_list", "gas_price")):
+        if all(_ in normalized_txn for _ in ("accessList", "gasPrice")):
             return self.chain.get_transaction_builder().new_unsigned_access_list_transaction(  # noqa: E501
-                **normalized_txn
+                **dict_keys_to_snake_case(normalized_txn)
```
fselmo, line 707: Yeah, I'm not convinced we can ignore my comment above. It leads to many calls to this method throughout the backend.

eth_tester/backends/pyevm/main.py
```diff
@@ -700,15 +702,18 @@ def _get_normalized_and_signed_evm_transaction(
...
             )
         elif all(_ in normalized_txn for _ in DYNAMIC_FEE_TRANSACTION_PARAMS):
-            return self.chain.get_transaction_builder().new_unsigned_dynamic_fee_transaction(  # noqa: E501
-                **normalized_txn
-            )
-        return self.chain.create_unsigned_transaction(**normalized_txn)
+            test = self.chain.get_transaction_builder()
+            txn = dict_keys_to_snake_case(normalized_txn)
```
fselmo, line 711: Here again.

eth_tester/backends/pyevm/main.py
```diff
@@ -700,15 +702,18 @@ def _get_normalized_and_signed_evm_transaction(
...
+            result = test.new_unsigned_dynamic_fee_transaction(**txn)  # noqa: E501
+            return result
+        return self.chain.create_unsigned_transaction(
+            **dict_keys_to_snake_case(normalized_txn)
```
fselmo, line 715: etc...

eth_tester/normalization/inbound.py
```diff
@@ -99,17 +99,17 @@ def _normalize_inbound_access_list(access_list):
 
 TRANSACTION_NORMALIZERS = {
     "type": identity,
-    "chain_id": identity,
+    "chainId": identity,
```
fselmo, line 102: I think we can probably pass a `compose` within the normalizers for dicts so that we pass all of them through the `dict_keys_to_snake_case` and avoid all of those multiple calls to that method randomly in the backend.

I believe this is what you had started to do with the outbound but they pass through the serializers first. So for outbound it makes more sense to handle it in the serializers since we manually build the objects there. For inbound, this would make sense to handle in the normalization step.

fselmo (commented):
@reedsa I think this looks good but imo the camel -> snake should happen during the normalization step. We end up having to re-normalize the transaction outside of that step all over the codebase if we don't do that.

I'm open to your thoughts there because if we really think we're going to replace everything with pydantic models, and I do believe we probably are, then we can avoid the double work and handle it in the pydantic model conversion PR.

Thoughts?

---

edit: as per the description on #318

eth_tester/normalization/inbound.py
reedsa: I see that eels already has it's [own normalizer](eth_tester/backends/eels/eels_normalizers.py) that converts everything to camelCase and stores that in the backend class. I think we can leave that as is (for now).

I understand the point of converting from snake to camel in the inbound normalizer, however I thought inbound calls should expect camel case like the JSON-RPC spec has. The validators check that the keys are camel cased and then there is nothing to normalize other than the values themselves.

I am not sure about the snake case as part of the backend internals. It appears that EELS does a secondary normalization and that is what I've replicated with the calls to `dict_keys_to_snake_case` in pyevm. Does that make sense?

fselmo: Yeah I guess EELS is different and probably the standard now. I think ideally, each backend should just have their own normalizer and perhaps have a global utils bucket they can share things from but that isn't the case and we're going to change this behavior anyhow.

Yeah we can leave this alone for sure. I'll mark all this as resolved.

fselmo (approved):
lgtm! good progress 📈

reedsa:
Looking forward to removing all of this complexity with these validators/normalizers. My main takeaway from this PR is the tests. Even though most of it will be removed I think they will be a useful point of reference when we get the new models in place.

PR #335, opened 2025-05-08
json-rpc server

reedsa:
### What was wrong?

Related to Issue #
Closes #

### How was it fixed?

Uses click, uvicorn and jsonrpc-py to start an jsonrpc server and serves the `get_accounts` endpoint.

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture


fselmo:
I sort of imagined this as a separate repository to *eth-tester*. I'm open to this idea but it does have a bit more separation of concerns and logic if we maintained a separate repository that uses *eth-tester* as a dependency instead. 

Curious about thoughts on that, cc: @kclowes @pacrob

---
edit: I started a very scrappy version of this [here](https://github.com/fselmo/fyke). But then I think we can actually write its own test suite to test the server logic on its own?

reedsa:
I'm cool with that too. Didnt get any push back when I mentioned it today, but this is the general idea for the server. You had started a separate packaged one too right?

This one uses jsonrpc-py, which seems like the most modern supported package. My original one uses Flask, https://github.com/reedsa/teth

fselmo:
It looks like main difference between `jsonrpc-py` and `jsonrpcserver` (what I used) is `jsonrpc-py` has client side also, not just server, and it supports websocket. I think the websocket support is nice to have but definitely should check temperature on maintenance. Seems like a good approach either way.

fselmo:
> I think the websocket support is nice to have but definitely should check temperature on maintenance. Seems like a good approach either way.

Yeah `jsonrpc-py` has the latest python version support, that looks like the way to go 👍🏼 

---
edit: Maybe another upside to maintaining as a separate repository is that `jsonrpc-py` only supports python 3.12 and up I think. So we can version control a bit differently since the server takes the main focus there.

PR #293, opened 2024-05-06
Upgrade eth-account

kclowes:
### What was wrong?
- eth-account moved `TypedTransactions` up a dir, from `eth_account._utils` to `eth_account`. 
- eth-account also changed `rawTransaction` -> `raw_transaction`.

### How was it fixed?

- Changed the `TypedTransactions` import path
- Changed `rawTransaction` -> `raw_transaction`
- Bumped eth-account to `>=0.12.3`

### Todo:

- [x] Clean up commit history
- [x] Add or update documentation related to these changes
- [x] Add entry to the [release notes](https://github.com/ethereum/eth-tester/blob/main/newsfragments/README.md)

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #340, opened 2025-05-19
Rename eth_sendRawTransaction to eth_send_raw_transaction

gap-editor:
Rename client binding `eth_sendRawTransaction` to `eth_send_raw_transaction` because we have renamed `rawTransaction` → `raw_transaction` (#293). I think this change is logical and it must be done with #293


kclowes:
The rpc method is eth_sendRawTransaction, so this isn't something we can change on the eth-tester side.

